PR #18, opened 2022-05-11
Fix build tags for `darwin` after refactoring of metrics handles

kvch:
With the refactoring of `SetupLinuxBSDFDMetrics`, `process.Stats` got imported on darwin. However, to use `process` package on darwin CGO is required. I have separated the process package so `process.Stats` can be imported on `darwin` without CGO. This way, the `SetupLinuxBSDFDMetrics` stub can be called when necessary.

Required by https://github.com/elastic/fleet-server/pull/1351

### DIFF
--- a/metric/system/process/process.go
+++ b/metric/system/process/process.go
@@ -15,107 +15,23 @@
 // specific language governing permissions and limitations
 // under the License.
 
-//go:build darwin || freebsd || linux || windows || aix
-// +build darwin freebsd linux windows aix
+//go:build (darwin && cgo) || freebsd || linux || windows || aix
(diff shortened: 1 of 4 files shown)


fearful-symmetry (approved):
Ran some tests on an M1, LGTM, but we may want to integrate some of the bugfixes from https://github.com/elastic/beats/pull/31595, or put that in a different PR.

kvch:
I will open a separate PR for the changes in https://github.com/elastic/beats/pull/31595

kvch:
Follow-up PR: https://github.com/elastic/elastic-agent-system-metrics/pull/19

PR #26, opened 2022-05-23
Fix issue with certain V1 cgroup events not reporting root IDs

fearful-symmetry:
## What does this PR do?

This is a migration of https://github.com/elastic/beats/pull/31671 , since the files got moved over here.

This is a fix for a bug in cgroups, where the new V2-compatible code wasn't properly filtering controller paths before trying to find the "root" controller ID. This also cleans up some of the cgroup testing code, including the testdata files, since it was a pain.

## Why is it important?
This is a bugfix

## Checklist


metric/system/cgroup/reader_test.go
```diff
@@ -33,8 +33,51 @@ import (
...
+// 	}
+
+// 	if vers == CgroupsV2 {
+// 		t.Logf("Got V2 process.")
+// 		stats, err := reader.GetV2StatsForProcess(pid)
+// 		assert.NoError(t, err, "error in GetV2StatsForProcess")
+// 		t.Logf("Event: %#v", stats)
+// 	}
+
+// }
```
belimawr, line 61: I believe we can remove those lines.

metric/system/cgroup/reader_test.go
```diff
@@ -33,8 +33,51 @@ import (
...
+
+func TestV1EventDifferentPaths(t *testing.T) {
+	pid := 3757
+	reader, err := NewReader(resolve.NewTestResolver("testdata/ubuntu1804"), true)
+	assert.NoError(t, err, "error in NewReader")
+
+	stats, err := reader.GetV1StatsForProcess(pid)
+	assert.NoError(t, err, "error in GetV1StatsForProcess")
```
belimawr, line 69: It's better to use `require` here so the test is interrupted. If we can't read/get the stats there is no point in continuing the tests.
```suggestion
	require.NoError(t, err, "error in NewReader")

	stats, err := reader.GetV1StatsForProcess(pid)
	require.NoError(t, err, "error in GetV1StatsForProcess")
```

metric/system/cgroup/reader_test.go
```diff
@@ -33,8 +33,51 @@ import (
...
+
+	if stats == nil {
+		t.Fatal("no cgroup stats found")
+	}
+
+	// Make sure we can handle root paths properly
+	assert.Equal(t, "/system.slice/networkd-dispatcher.service", stats.Path)
+	assert.Equal(t, "networkd-dispatcher.service", stats.ID)
+
```
belimawr, line 78:
```suggestion
```

metric/system/cgroup/reader_test.go
```diff
@@ -82,10 +125,18 @@ func TestReaderGetStatsV2(t *testing.T) {
...
 	require.NotZero(t, stats.CPU.Stats.Usage.NS)
 	require.NotZero(t, stats.Memory.Mem.Usage.Bytes)
 	require.NotZero(t, stats.IO.Pressure["some"].Sixty.Pct)
+	json, err := json.MarshalIndent(stats, "", "  ")
+	if err != nil {
+		t.Fatal(err)
+	}
 
+	t.Log(string(json))
```
belimawr, line 139: I believe we don't need this log anymore, right?
```suggestion
```

metric/system/cgroup/reader_test.go
```diff
@@ -82,10 +125,18 @@ func TestReaderGetStatsV2(t *testing.T) {
 	require.NotNil(t, stats.Memory)
 	require.NotNil(t, stats.IO)
 
+	assert.Equal(t, pathv2, stats.Path)
+	assert.Equal(t, idv2, stats.ID)
```
belimawr, line 129: It seems the use of `assert` and `require` on this test is reversed. `require` will stop the test execution while `require` will keep going and logging all errors.

I believe the first two:
```go
	reader, err := NewReader(resolve.NewTestResolver("testdata/docker"), true)
	assert.NoError(t, err, "error in NewReader")

	stats, err := reader.GetV2StatsForProcess(312)
	assert.NoError(t, err, "error in GetV2StatsForProcess")
```
Should be `require` so the tests stops in case of failure, then we can just assert the rest.

belimawr (changes requested):

fearful-symmetry:
@belimawr / @leehinman looks like git wants a code owner to check this, although I'm not sure why @kvch / myself aren't in said code owner group.

cmacknz:
> @belimawr / @leehinman looks like git wants a code owner to check this, although I'm not sure why @kvch / myself aren't in said code owner group.

The code owner auto assigns the data plane team, and we have github configured to randomly choose two reviewers from the team. You can manually assign the reviewers when you create a PR and it will override this as long as they satisfy the code owner criteria.

leehinman (approved):
LGTM

PR #27, opened 2022-05-24
Move pacakges from Metricbeat: `internal/metrics/cpu` and `internal/metrics/memory`

kvch:
## What does this PR do?

This PR moves packages from `metricbeat/internal/system` as we agreed in https://github.com/elastic/beats/pull/31615#issuecomment-1125343951

The first commit copies the files from Metricbeat, the second one (https://github.com/elastic/elastic-agent-system-metrics/commit/1cd2f259983354fde0901c0f716a4bb4356dd7ba) contains my fixes.

I did not move `sysinit` because it is heavily tied to Metricbeat code.

## Why is it important?

Move system related code to one package so it can be reused in multiple places.

## Checklist


fearful-symmetry (approved):
No clue what's up with the linter error, changes LGTM.

PR #37, opened 2022-06-29
Move registry used by handles processor

fearful-symmetry:
## What does this PR do?

Fix for a bug reported by @cmacknz where the migration process of metrics from `libbeat` to here also changed the location of the `handles` monitoring metrics. This changes it back from `"system"` to `"beat"`. We don't have any tests here, so a tad hard to validate this until it makes its way up to the beats repo again.

## Checklist


cmacknz (approved):
We'll want to update the dependency in the branch for https://github.com/elastic/beats/pull/32153, the tests there reproduce the problem.

PR #40, opened 2022-07-06
Add IsZero to CPU Totals, report value for monitoring

fearful-symmetry:
## What does this PR do?

This fixes three issues:

- Gives `CPUTotals` an `IsZero` method, so we don't report `{}` fields during certain events
- Fills out the `Value` field for `GetSelf()`, which is used by beats self-monitoring, and apparently that value is used by the `elastic_agent` dashboard.
- Make processes fetched with `GetSelf()` persist across calls

### DIFF
--- a/metric/system/process/process.go
+++ b/metric/system/process/process.go
@@ -133,6 +133,8 @@ func (procStats *Stats) GetSelf() (ProcState, error) {
 	if err != nil {
 		return ProcState{}, fmt.Errorf("error fetching PID %d: %w", self, err)
 	}
+	// Copy Value over here, which would normally happen in system/process as part of percent calculation
+	pidStat.CPU.Total.Value = opt.FloatWith(float64(pidStat.CPU.Total.Ticks.ValueOr(0)))
 
 	return pidStat, nil
 }
--- a/metric/system/process/process.go
+++ b/metric/system/process/process.go
@@ -30,6 +30,7 @@ import (
 	"github.com/elastic/elastic-agent-libs/mapstr"
 	"github.com/elastic/elastic-agent-libs/opt"
 	"github.com/elastic/elastic-agent-libs/transform/typeconv"
+	"github.com/elastic/elastic-agent-system-metrics/metric"
 	"github.com/elastic/elastic-agent-system-metrics/metric/system/resolve"
 )
 
@@ -133,8 +134,6 @@ func (procStats *Stats) GetSelf() (ProcState, error) {
 	if err != nil {
 		return ProcState{}, fmt.Errorf("error fetching PID %d: %w", self, err)
 	}
-	// Copy Value over here, which would normally happen in system/process as part of percent calculation
-	pidStat.CPU.Total.Value = opt.FloatWith(float64(pidStat.CPU.Total.Ticks.ValueOr(0)))
 
 	return pidStat, nil
 }
@@ -204,6 +203,9 @@ func (procStats *Stats) pidFill(pid int, filter bool) (ProcState, bool, error) {
 		}
 	} // end cgroups processor
 
+	if status.CPU.Total.Ticks.Exists() {
+		status.CPU.Total.Value = opt.FloatWith(metric.Round(float64(status.CPU.Total.Ticks.ValueOr(0))))
+	}
 	if ok {
 		status = GetProcCPUPercentage(last, status)
 	}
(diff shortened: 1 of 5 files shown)


metric/system/process/process.go
```diff
@@ -133,6 +133,8 @@ func (procStats *Stats) GetSelf() (ProcState, error) {
 	if err != nil {
 		return ProcState{}, fmt.Errorf("error fetching PID %d: %w", self, err)
 	}
+	// Copy Value over here, which would normally happen in system/process as part of percent calculation
+	pidStat.CPU.Total.Value = opt.FloatWith(float64(pidStat.CPU.Total.Ticks.ValueOr(0)))
```
cmacknz, line 137: Should or could this be fixed in https://github.com/elastic/integrations/tree/main/packages/elastic_agent instead?

metric/system/process/process_types.go
```diff
@@ -103,6 +103,10 @@ type ProcLimits struct {
 
 // Implementations
 
+func (t CPUTotal) IsZero() bool {
```
cmacknz, line 106: Where does this get used? Is there an interface somewhere with this method?

metric/system/process/process.go
cmacknz: Also fixing it in the integration or dashboard itself means we aren't coupled to the stack release.

metric/system/process/process_types.go
fearful-symmetry: Yes. It's used by the `go-structform` code to check to see if a struct should be removed from whatever JSON event is being generated.

metric/system/process/process.go
fearful-symmetry: Yah, currently on the fence about how to fix this, digging into the issue a bit more, since I think there's other things at work.

fearful-symmetry:
Okay, tinkered with this a bit more. The logic that adds `value` the way it does is ancient, and I don't fully understand why it's like that, but for now I've just decided to add the `value` metric in a way that...seems more reasonable?

metric/system/process/process_test.go
```diff
@@ -118,6 +119,22 @@ func TestProcessList(t *testing.T) {
...
 
+func TestSelfPersist(t *testing.T) {
+	stat, err := initTestResolver()
+	require.NoError(t, err, "Init()")
+	first, err := stat.GetSelf()
+	require.NoError(t, err, "First GetSelf()")
+
+	// The first process fetch shouldn't have percentages, since we don't have >1 procs to compare
+	assert.False(t, first.CPU.Total.Pct.Exists(), "total.pct should not exist")
+	time.Sleep(time.Millisecond * 5)
```
cmacknz, line 130: Why 5 ms? Why is the delay here necessary?

metric/system/process/process.go
```diff
@@ -133,7 +134,7 @@ func (procStats *Stats) GetSelf() (ProcState, error) {
 	if err != nil {
 		return ProcState{}, fmt.Errorf("error fetching PID %d: %w", self, err)
 	}
-
+	procStats.ProcsMap[self] = pidStat
```
cmacknz, line 137: What is this doing? I assume it is related to this in the PR description:

> Make processes fetched with GetSelf() persist across calls

metric/system/process/process_test.go
fearful-symmetry: Interestingly enough, it depends on the OS. On darwin and windows, CI will fail because there's no time delta between the two measurements. The actual count was largely arbitrary.

metric/system/process/process_test.go
```diff
@@ -118,6 +119,22 @@ func TestProcessList(t *testing.T) {
 	}
 }
 
+func TestSelfPersist(t *testing.T) {
+	stat, err := initTestResolver()
+	require.NoError(t, err, "Init()")
+	first, err := stat.GetSelf()
+	require.NoError(t, err, "First GetSelf()")
+
+	// The first process fetch shouldn't have percentages, since we don't have >1 procs to compare
```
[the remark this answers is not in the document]
fearful-symmetry: The code uses the delta between two measurements to calculate percentages. Therefore, the very first report we get will never have any percentages, as it will always save off the "last" value for calculation.

metric/system/process/process.go
fearful-symmetry: Yep. We need to save off the previous value in order to calculate CPU usage percentages.

metric/system/process/process_test.go
cmacknz: Just add a comment explaining what this is for then, the only other way I could see to have this work would be to mock time.Now() or something to control the time between calls.

cmacknz:
1) Add a changelog entry for this and create a release+tag for this change.
2) As a follow up create a beats PR updating this package, and then backport it to the 8.3 branch. There is going to be an 8.3.3 release this fix can go out in.

cmacknz (approved):

PR #43, opened 2022-07-22
fix thread safety in process code

fearful-symmetry:
## What does this PR do?

This fixes https://github.com/elastic/beats/issues/32467

This code originates from metricbeat, which never worried too much about threadsafety, since a single metricset just lives in its own thread. However, the code has carried over to beat's self-monitoring in `report/`. A previous PR fixed a bug where `GetSelf()` would not calculate percentages, as it never accessed its own internal map of processes that it uses to calculate percentages. However, `setup.go` was creating a single instance of the process `Stat` struct to send to multiple function callbacks in the monitoring, hence a concurrent map access. 

Wanted to get this PR up fast, still testing on non-linux systems.

## Why is it important?

Potential concurrent hashmap access issue.

## Checklist

- [x] Test on darwin
- [ ] Test on Windows

### DIFF
--- a/metric/system/process/process_common.go
+++ b/metric/system/process/process_common.go
@@ -23,6 +23,7 @@ package process
 import (
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/elastic/elastic-agent-libs/logp"
 	"github.com/elastic/elastic-agent-libs/match"
@@ -33,8 +34,42 @@ import (
 	sysinfo "github.com/elastic/go-sysinfo"
 )
 
-// ProcsMap is a map where the keys are the names of processes and the value is the Process with that name
+//ProcsMap is aconvinence wrapper type
 type ProcsMap map[int]ProcState
+
+// ProcsTrack is a map where the keys are the names of processes and the value is the Process with that name
+//type ProcsTrack map[int]ProcState
+type ProcsTrack struct {
+	pids ProcsMap
+	mut  *sync.Mutex
+}
+
+func NewProcsMap() *ProcsTrack {
+	return &ProcsTrack{
+		pids: make(ProcsMap, 0),
+		mut:  &sync.Mutex{},
+	}
+}
+
+func (pm *ProcsTrack) GetPid(pid int) (ProcState, bool) {
+	pm.mut.Lock()
+	defer pm.mut.Unlock()
+	proc, ok := pm.pids[pid]
+	return proc, ok
+}
+
+func (pm *ProcsTrack) SetPid(pid int, ps ProcState) {
+	pm.mut.Lock()
+	defer pm.mut.Unlock()
+	pm.pids[pid] = ps
+}
+
+func (pm *ProcsTrack) SetMap(pids map[int]ProcState) {
+	pm.mut.Lock()
+	defer pm.mut.Unlock()
+	pm.pids = pids
+
+}
 
 // ProcCallback is a function that FetchPid* methods can call at various points to do OS-agnostic processing
 type ProcCallback func(in ProcState) (ProcState, error)
@@ -53,7 +88,7 @@ type CgroupPctStats struct {
 type Stats struct {
 	Hostfs        resolve.Resolver
 	Procs         []string
-	ProcsMap      ProcsMap
+	ProcsMap      *ProcsTrack
 	CPUTicks      bool
 	EnvWhitelist  []string
 	CacheCmdLine  bool
@@ -128,7 +163,7 @@ func (procStats *Stats) Init() error {
 		procStats.Hostfs = resolve.NewTestResolver("/")
 	}
 
-	procStats.ProcsMap = make(ProcsMap)
+	procStats.ProcsMap = NewProcsMap()
 
 	if len(procStats.Procs) == 0 {
 		return nil
--- a/metric/system/process/process_common.go
+++ b/metric/system/process/process_common.go
@@ -34,20 +34,19 @@ import (
 	sysinfo "github.com/elastic/go-sysinfo"
 )
 
-//ProcsMap is aconvinence wrapper type
+//ProcsMap is a convinence wrapper for the oft-used ideom of map[int]ProcState
 type ProcsMap map[int]ProcState
 
-// ProcsTrack is a map where the keys are the names of processes and the value is the Process with that name
-//type ProcsTrack map[int]ProcState
+// ProcsTrack is a thread-safe wrapper for a process Stat object's internal map of processes.
 type ProcsTrack struct {
 	pids ProcsMap
-	mut  *sync.Mutex
+	mut  sync.Mutex
 }
 
 func NewProcsMap() *ProcsTrack {
 	return &ProcsTrack{
 		pids: make(ProcsMap, 0),
-		mut:  &sync.Mutex{},
+		mut:  sync.Mutex{},
 	}
 }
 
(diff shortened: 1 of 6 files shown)


cmacknz:
To unblock 8.3.3 as safely as possible we are going to downgrade the system metrics package in affected versions: https://github.com/elastic/beats/issues/32467#issuecomment-1192799872

We will merge this to beats main and 8.3 after the 8.3.3 release completes (really after the next build candidate is generated).

metric/system/process/process_common.go
```diff
@@ -33,9 +34,43 @@ import (
 	sysinfo "github.com/elastic/go-sysinfo"
 )
 
-// ProcsMap is a map where the keys are the names of processes and the value is the Process with that name
+//ProcsMap is aconvinence wrapper type
```
cmacknz, line 37:
```suggestion
//ProcsMap is a convinence wrapper type
```

metric/system/process/process_common.go
```diff
@@ -33,9 +34,43 @@ import (
...
 type ProcsMap map[int]ProcState
 
+// ProcsTrack is a map where the keys are the names of processes and the value is the Process with that name
+//type ProcsTrack map[int]ProcState
+type ProcsTrack struct {
+	pids ProcsMap
+	mut  *sync.Mutex
```
cmacknz, line 44: Why is this a pointer? I'm not sure it needs to be one. The zero value is safe for use.

cmacknz:
Is there a test we can write to prove this is thread safe? Maybe the test we need should be added in beats?

fearful-symmetry:
> Is there a test we can write to prove this is thread safe? Maybe the test we need should be added in beats?

@cmacknz  Yah, I was wondering about that. If anything, I think the test should go in the `report` library, since it doesn't have any tests, and that's where the threaded code actually originated.

cmacknz:
> Yah, I was wondering about that. If anything, I think the test should go in the report library, since it doesn't have any tests, and that's where the threaded code actually originated.

Should we just document that this code isn't thread safe and put the locking in beats? I haven't looked at that code yet so I don't know how hard it is, but it seems a bit weird for the fix and the test to live in two different places.

fearful-symmetry:
> Should we just document that this code isn't thread safe and put the locking in beats?

Well, theoretically it should be threadsafe now. The reporting/monitoring code that actually creates and spins up per-thread callbacks are in `setup.go` and `elastic-agent-libs/reporting`, so any further thread safety fixes would have to go there. 

If we wanted to be extra paranoid, we could have just have each monitoring callback function have its own instance of `Stat`, but if we're confident this code is thread safe that seems unnecessary.

fearful-symmetry:
FWIW, a test on the monitoring code would almost be more an integration test, since we're hitting both the monitoring subsystem and any metrics code attached to it.

fearful-symmetry:
Added a small test to the report library. System metrics like this are always a tad awkward to write tests for, since a lot of metrics will just be context/OS dependent.

metric/system/process/process_common.go
```diff
@@ -33,9 +34,42 @@ import (
...
 )
 
-// ProcsMap is a map where the keys are the names of processes and the value is the Process with that name
+//ProcsMap is a convinence wrapper for the oft-used ideom of map[int]ProcState
 type ProcsMap map[int]ProcState
 
+// ProcsTrack is a thread-safe wrapper for a process Stat object's internal map of processes.
+type ProcsTrack struct {
+	pids ProcsMap
+	mut  sync.Mutex
```
cmacknz, line 43: This could be a `RWMutex` instead to allow multiple concurrent calls to `GetPid`.

metric/system/process/process_common.go
```diff
@@ -33,9 +34,42 @@ import (
...
+}
+
+func NewProcsMap() *ProcsTrack {
+	return &ProcsTrack{
+		pids: make(ProcsMap, 0),
+		mut:  sync.Mutex{},
```
cmacknz, line 49:
```suggestion
```
I don't think this is necessary, `mut sync.Mutex` should be usable immediately since it is equivalently default constructed.

metric/system/process/process_common.go
```diff
@@ -33,9 +34,42 @@ import (
...
+//ProcsMap is a convinence wrapper for the oft-used ideom of map[int]ProcState
 type ProcsMap map[int]ProcState
 
+// ProcsTrack is a thread-safe wrapper for a process Stat object's internal map of processes.
+type ProcsTrack struct {
+	pids ProcsMap
+	mut  sync.Mutex
+}
+
+func NewProcsMap() *ProcsTrack {
```
cmacknz, line 46:
```suggestion
func NewProcsTrack() *ProcsTrack {
```

nit: The constructor name should match the type name.

cmacknz: Similarly I see the receiver is named `pm` in the methods below, did you switch the name from `ProcsMap` to `ProcsTrack` and only partially update the naming?

report/metrics_report_test.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		if key == "memstats.rss" {
+			gotMem = true
+		}
+	}
+
+	//iterate over the processes a few times,
+	// with the concurrency (hopefully) emulating what might
+	// happen if this was an HTTP endpoint getting multiple GET requests
+	iter := 5
```
cmacknz, line 53: If you remove the mutex from `ProcsTrack` this test will still pass. If you increase the concurrency, and add a broadcast channel close to wake all the goroutines at the same time it will fail within 3-5 test runs reliable with the mutex removed:

```go
diff --git a/report/metrics_report_test.go b/report/metrics_report_test.go
index 0ede44f39..207b261f0 100644
--- a/report/metrics_report_test.go
+++ b/report/metrics_report_test.go
@@ -50,15 +50,18 @@ func TestSystemMetricsReport(t *testing.T) {
        //iterate over the processes a few times,
        // with the concurrency (hopefully) emulating what might
        // happen if this was an HTTP endpoint getting multiple GET requests
-       iter := 5
+       iter := 100
        var wait sync.WaitGroup
        wait.Add(iter)
+       ch := make(chan struct{})
        for i := 0; i < iter; i++ {
                go func() {
+                       <-ch
                        processMetrics.Do(monitoring.Full, testFunc)
                        wait.Done()
                }()
        }
+       close(ch)

        wait.Wait()
        assert.True(t, gotCPU, "Didn't find cpu.total.ticks")
```

Without synchronizing the goroutines they probably just run as they are started and complete too quickly to hit the race condition.

cmacknz:
Needs a CHANGELOG entry :)

report/metrics_report_test.go
fearful-symmetry: Ohhh, interesting, thanks.

cmacknz (approved):

LeeDr:
> [fearful-symmetry](https://github.com/fearful-symmetry) merged commit [c8f77ef](https://github.com/elastic/elastic-agent-system-metrics/commit/c8f77efb8c7700a27278c485b0ffc985e6c715b8) into elastic:main [17 days ago](https://github.com/elastic/elastic-agent-system-metrics/pull/43#event-7058826751)

This commit may have been made to the 8.3 branch, but it was made well after 8.3.0 was released so I think this should have made it into the 8.3.3 release and should be labeled with that.

PR #53, opened 2022-09-26
Add Process State API

fearful-symmetry:
## What does this PR do?

This is part of https://github.com/elastic/beats/pull/33169

Specifically, This adds a public API that reports the state of a given PID, and a special error if a given process does not exist.

This is a bit of an awkward fix, since there's no "good" place for this API to go; This library is mainly for handling the complexities of the `system/process` dataset, but it's only the library we have that reports process state. `go-sysinfo` might be a better fit for this, but there's no code for fetching process state there. So, for now, this is kind of the best we can do.

## Why is it important?

This is needed by https://github.com/elastic/beats/pull/33169 so we can distinguish between zombie processes and other process types.

## Checklist

### DIFF
--- a/metric/system/process/process.go
+++ b/metric/system/process/process.go
@@ -21,11 +21,14 @@
 package process
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"sort"
 	"strings"
 	"time"
+
+	psutil "github.com/shirou/gopsutil/process"
 
 	"github.com/elastic/elastic-agent-libs/mapstr"
 	"github.com/elastic/elastic-agent-libs/opt"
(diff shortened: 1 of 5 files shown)


metric/system/process/process.go
```diff
@@ -56,6 +59,27 @@ func ListStates(hostfs resolve.Resolver) ([]ProcState, error) {
...
 }
 
+// GetPIDState returns the state of a given PID
+// It will return ProcNotExist if the process was not found.
+func GetPIDState(hostfs resolve.Resolver, pid int) (PidState, error) {
+	// This library still doesn't have a good cross-platform way to distinguish between "does not eixst" and other process errors.
+	// This is a fairly difficult problem to solve in a cross-platform way
+	exists, err := psutil.PidExistsWithContext(context.Background(), int32(pid))
+	if err != nil {
+		return "", fmt.Errorf("Error truing to find process: %d: %w", pid, err)
```
leehinman, line 69:
```suggestion
		return "", fmt.Errorf("error trying to find process: %d: %w", pid, err)
```

metric/system/process/process_common.go
```diff
@@ -34,6 +34,9 @@ import (
 	sysinfo "github.com/elastic/go-sysinfo"
 )
 
+// ProcNotExist indicates that a process was not found.
+var ProcNotExist = errors.New("Process does not exist")
```
leehinman, line 38:
```suggestion
var ProcNotExist = errors.New("process does not exist")
```

leehinman (approved):
LGTM.  couple spelling and capitalization suggestions

PR #62, opened 2022-11-10
Add per-process network stats

fearful-symmetry:
## What does this PR do?
This is one half of https://github.com/elastic/beats/issues/7461

This adds per-process network stats on linux, taken from `/proc/PID/net/netstat` and `/proc/PID/net/SNMP`. 
These metrics are the same kind used in the docker module for reporting docker network summary data.

This also includes a filter function, since all the data in the aforementioned two procfs files includes a prodigious amount of metrics.

There's a notable limitation here, which is that the procfs handler we're importing from `go-sysinfo` can't handle an alternate procfs.

## Why is it important?

This is an asked-for feature.

## Checklist

### DIFF
--- a/metric/system/network/helpers.go
+++ b/metric/system/network/helpers.go
@@ -22,30 +22,52 @@ import (
 	sysinfotypes "github.com/elastic/go-sysinfo/types"
 )
 
+// MapProcNetCountersWithFilter converts the NetworkCountersInfo to a formatted mapstring,
+// and applies a filter to the resulting map. The filter should be an array key values, taken from /proc/PID/net/snmp or /proc/PID/net/netstat
+func MapProcNetCountersWithFilter(raw *sysinfotypes.NetworkCountersInfo, filter []string) mapstr.M {
+	return createMap(raw, filter)
+}
+
 // MapProcNetCounters converts the NetworkCountersInfo struct into a MapStr acceptable for sending upstream
 func MapProcNetCounters(raw *sysinfotypes.NetworkCountersInfo) mapstr.M {
+	return createMap(raw, []string{"all"})
+}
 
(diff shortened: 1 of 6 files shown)


metric/system/network/helpers.go
```diff
@@ -22,30 +22,52 @@ import (
...
-	}
-	for k, v := range map2 {
-		compMap[k] = checkMaxConn(k, v)
+	if len(filter) == 0 || filter[0] == "all" {
+		for k, v := range map1 {
+			compMap[k] = checkMaxConn(k, v)
+		}
+		for k, v := range map2 {
+			compMap[k] = checkMaxConn(k, v)
+		}
```
rdner, line 58: what if 2 maps have the same key, do we check twice?

metric/system/network/helpers.go
```diff
@@ -22,30 +22,52 @@ import (
...
+	} else {
+		for _, key := range filter {
+			if value, ok := map1[key]; ok {
+				compMap[key] = checkMaxConn(key, value)
+			}
+			if value, ok := map2[key]; ok {
+				compMap[key] = checkMaxConn(key, value)
+			}
```
rdner, line 66: Same comment here.

metric/system/network/helpers.go
fearful-symmetry: So, these values are only being merged across a given protocol, and in my experience, the kernel won't report the same metric for a given protocol twice, so it should be fine? Not sure how we would recover from that anyway, since it would seemingly be the same metric reported in two different places?

rdner: I just realised that `checkMaxConn` would just do a type cast, it's not really a big deal to run it twice for the same key. So, the current logic is that the second map has a priority over the first if the keys collide, which I think is fine.

rdner (approved):

PR #77, opened 2023-03-10
Fix io.stat parsing errors

fearful-symmetry:
## What does this PR do?

Closes https://github.com/elastic/beats/issues/32345

This makes the parsing of `io.stat` cgroup files _much_ more defensive; and also skips over errors that can happen while looking for device names.

## Why is it important?

This is causing lots of non-fatal yet annoying errors.

## Checklist

- Relates https://github.com/elastic/beats/issues/32345


cmacknz (approved):
LGTM, although not super familiar with the cgroup metrics format.

I did check this out and confirm that parsing the new .stat files you added fails with the previous version of the code. 

Remember to update every repository using this package with the fix. From the set of repos I have locally you need to update all three of Beats, Elastic Agent, and Fleet Server to fix this error everywhere.

```sh
❯ rg 'elastic-agent-system-metrics' -g 'go.mod' ../
../fleet-server/go.mod
11:     github.com/elastic/elastic-agent-system-metrics v0.4.4

../beats/go.mod
200:    github.com/elastic/elastic-agent-system-metrics v0.4.4

../elastic-agent/go.mod
18:     github.com/elastic/elastic-agent-system-metrics v0.4.4
```

cmacknz:
Elastic agent and fleet-server will still have this bug, they are using the previous v0.4.4 version of the package:

- https://github.com/elastic/fleet-server/blob/b67d3a0031201d36565aa24e49060821bccb7335/go.mod#L11
- https://github.com/elastic/elastic-agent/blob/main/go.mod#L18

cmacknz:
This should be fixed in the elastic agent process itself in the upcoming 8.7.1 release.

ebuildy:
I confirm, this is OK from v8.7.1, nice work!

jigar82:
@ebuildy  Is it added in 8.7.1 only reason for asking its not listed in release notes. 
https://www.elastic.co/guide/en/beats/libbeat/8.7/release-notes-8.7.1.html and PR has 8.8 candidate label https://github.com/elastic/beats/issues/32345
I have customer waiting so just wanted to pass info if this is indeed release in 8.7.1

ebuildy:
yes that I said, **I can confirm this is working from v8.7.1**

PR #80, opened 2023-03-27
Add user data to monitoring setup

fearful-symmetry:
## What does this PR do?

This is a simple PR that migrates the code that adds user data to the monitoring registry from beats to here. The original code is in `libbeat/cmd/instance/beat.go`, and needs to be used in the shipper in order to bring the shipper up to feature parity with the beats in terms of monitoring. Just copy-and-pasting the user metrics is a bit ugly, so I figured I'd be better off migrating the metrics.

## Why is it important?

We want to reduce code duplication across beats.

## Checklist

### DIFF
--- a/report/setup.go
+++ b/report/setup.go
@@ -22,7 +22,10 @@ package report
 
 import (
 	"fmt"
+	"os"
+	"os/user"
 	"runtime"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmacknz (approved):
Make sure we remove the duplicated code from Beats if we can.

PR #84, opened 2023-04-05
Bumping up version for go-sysinfo dependency

ycombinator:
## What does this PR do?

This PR bumps up the version of the `go-sysinfo` dependency.

## Why is it important?

To avoid using pseudo-version numbers.


blakerouse (approved):
Looks good.

PR #91, opened 2023-07-03
migrate pipeline to buildkite

sharbuz:
## What does this PR do?

Migration of the pipeline from the Jenkins to the Buildkite

## Checklist

- [x] Lint
- [x] Test

## Related issues

-

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,15 +11,17 @@ steps:
     key: lint-test
     command: ".buildkite/scripts/lint_test.sh"
     agents:
-      provider: "gcp"
       image: "${LINUX_AGENT_IMAGE}"
+      cpu: "8"
+      memory: "4G"
 
   - label: ":linux: Tests on Linux"
     key: linux-test
     command: ".buildkite/scripts/run-linux-tests.sh"
     agents:
-      provider: "gcp"
       image: "${LINUX_AGENT_IMAGE}"
+      cpu: "8"
+      memory: "4G"
     artifact_paths:
       - "build/*.xml"
 
@@ -38,7 +40,9 @@ steps:
           artifacts: "build/*.xml"
           fail-build-on-error: true
     agents:
-      provider: "gcp"
+      image: "${LINUX_AGENT_IMAGE}"
+      cpu: "4"
+      memory: "2G"
     depends_on:
       - step: "linux-test"
         allow_failure: true
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,8 +2,7 @@
 
 env:
   SETUP_GVM_VERSION: 'v0.5.0'
-  GO_VERSION: "1.19.10"
-  GO_VERSION_CHOCO: "1.19.9"
+  GO_VERSION_CHOCO: "1.20.5"
   LINUX_AGENT_IMAGE: "golang:${GO_VERSION}"
   WINDOWS_AGENT_IMAGE: "family/ci-windows-2022"
 
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -33,12 +33,12 @@ steps:
       provider: "gcp"
       image: "${WINDOWS_AGENT_IMAGE}"
     artifact_paths:
-      - "build/*.xml"
+      - "*.xml"
 
   - label: ":junit: Junit annotate"
     plugins:
       - junit-annotate#v2.4.1:
-          artifacts: "build/*.xml"
+          artifacts: "*.xml"
           fail-build-on-error: true
     agents:
       image: "${LINUX_AGENT_IMAGE}"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,40 +7,40 @@ env:
   WINDOWS_AGENT_IMAGE: "family/ci-windows-2022"
 
   steps:
-  - label: ":golangci-lint: Lint"
-    key: lint-test
-    command: ".buildkite/scripts/lint_test.sh"
-    agents:
-      provider: "gcp"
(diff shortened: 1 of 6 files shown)


sharbuz:
buildkite test it

jlind23:
buildkite test this

.buildkite/scripts/common.sh
```diff
@@ -0,0 +1,34 @@
...
+
+create_workspace() {
+    if [[ ! -d "${WORKSPACE}" ]]; then
+    mkdir -p ${WORKSPACE}
+    fi
+}
+
+with_go() {
+    echo "Setting up the Go environment..."
+    retry 5 curl -sL -o ${WORKSPACE}/gvm "https://github.com/andrewkroh/gvm/releases/download/${SETUP_GVM_VERSION}/gvm-linux-amd64"
```
mrodm, line 15: Missing the definition of this retry function.
Should this function be added to this script ?

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,46 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 
-steps:
-  - label: "Example test"
-    command: echo "Hello!"
+env:
```
mrodm, line 3: Missing to add in this PR the `.buildkite/post-checkout` hook.

That hook needs to be added in all repositories for now, so the builds triggered in Pull Requests run the commit resulting of merging the target branch with the head of the Pull Request.

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,46 @@
...
+  LINUX_AGENT_IMAGE: "family/core-ubuntu-2004"
+  WINDOWS_AGENT_IMAGE: "family/ci-windows-2022"
+
+  steps:
+  - label: ":golangci-lint: Lint"
+    key: lint-test
+    command: ".buildkite/scripts/lint_test.sh"
+    agents:
+      provider: "gcp"
+      image: "${LINUX_AGENT_IMAGE}"
```
mrodm, line 15: Do these steps require docker to run their tests/lint? 
Could it be used just the golang image for these steps ?

jlind23:
buildkite test this

jlind23:
buildkite test this

rdner (commented):
I assume this PR is supposed to migrate this https://github.com/elastic/elastic-agent-system-metrics/blob/8b297f79fade6dfb664a9a1a51e25214a12996cd/.ci/Jenkinsfile#L39-L49

Why can't we just keep the mage target and setting up mage instead?

.buildkite/pipeline.yml
sharbuz: added

rdner:
@jlind23 if so, what are these files then?

* .buildkite/scripts/run-linux-tests.sh
* .buildkite/scripts/run-win-tests.ps1

I thought the whole point of using mage was to avoid OS specific scripting.

All these files definitely need more comments that explain their purpose.

rdner:
Ideally, we should only install Go and mage, the rest should go into mage targets.

jlind23:
@sharbuz @mrodm correct me if I am wrong please.
Scripts are executed on a platform basis, each OS will have its own buildekite agent that will agent the pipeline.
These scripts are not only running the mage target as they are also settings up the agent environment.

mrodm:
> @sharbuz @mrodm correct me if I am wrong please. Scripts are executed on a platform basis, each OS will have its own buildekite agent that will agent the pipeline. These scripts are not only running the mage target as they are also settings up the agent environment.

Exactly, buildkite agents have minimal tools installed on it. And therefore, the aim of those scripts are to install everything needed (gvm, golang, mage, some golang libs, docker-compose if needed...)

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,51 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 
+env:
+  SETUP_GVM_VERSION: 'v0.5.0'
+  GO_VERSION: "1.19.10"
+  GO_VERSION_CHOCO: "1.20.1"
```
rdner, line 6: Is it possible to take Go version from https://github.com/elastic/elastic-agent-system-metrics/blob/main/.go-version ?

It's very important to maintain exactly the same version of Go as specified in this file. We have automation that updates the version when we're ready for the transition, some of minor version updates require us to make some changes to the code.

sharbuz: It just tests because I don't see the 1.19.10 version as supported here: https://community.chocolatey.org/packages/golang#versionhistory
Also, I asked about it below on that page...

mrodm (approved):
LGTM

.buildkite/pipeline.yml
sharbuz: I used this approach:
https://github.com/elastic/elastic-agent-system-metrics/blob/cf8c16fbbbb4aaf30c635f693702ffcb30490bc3/.buildkite/hooks/post-checkout#L5
For Windows tests, I used the 1.20.5 version of Golang. 
This line should be changed after migration of the repository to Go v.1.20:
https://github.com/elastic/elastic-agent-system-metrics/blob/cf8c16fbbbb4aaf30c635f693702ffcb30490bc3/.buildkite/scripts/run-win-tests.ps1#L33
to:
`withGolang $env:GO_VERSION`
and remove this one: 
https://github.com/elastic/elastic-agent-system-metrics/blob/cf8c16fbbbb4aaf30c635f693702ffcb30490bc3/.buildkite/pipeline.yml#L5

sharbuz:
@rdner Do you have any more questions? Can I merge this PR?

rdner (approved):

rdner:
@sharbuz no, I think we've clarified all of them.

PR #96, opened 2023-07-28
`process.GetInfoForPid` collects the total thread count

AndersonQ:
## What does this PR do?
Collect the total thread count, num_threads, as part of the information collected by `process.GetInfoForPid`

* On linux:
  -  it extends the information parsed from /proc/[PID]/stat to get the total thread count
  -  unit tests added, they mock /proc/[PID]/stat
* On darwin:
  - it uses the syscall proc_pidinfo
  - unit test only validates GetInfoForPid does not return an error and the total thread count ins't zero
* On windows:
  - it uses the Process Snapshotting APIs: https://learn.microsoft.com/en-us/previous-versions/windows/desktop/proc_snap/overview-of-process-snapshotting
  - unit test only validates GetInfoForPid does not return an error and the total thread count ins't 
  - there is a manual test using a prebuilt C++ binary, its src is included, which calls the same windows APIs so the returned total thread count can be compared against the Go implementation.

## Why is it important?

To enable metricbeat to collect the total thread count per process as requested by https://github.com/elastic/beats/issues/34461.

## Checklist

## Related issues

- Relates to https://github.com/elastic/beats/issues/34461


metric/system/process/syscall_windows.go
```diff
@@ -0,0 +1,78 @@
...
+// PssCaptureSnapshot docs in https://learn.microsoft.com/en-us/windows/win32/api/processsnapshot/nf-processsnapshot-psscapturesnapshot
+// PssQuerySnapshot docs in https://learn.microsoft.com/en-us/windows/win32/api/processsnapshot/nf-processsnapshot-pssquerysnapshot
+
+// Use golang.org/x/sys/windows/mkwinsyscall instead of adriansr/mksyscall
+// below once https://github.com/golang/go/issues/42373 is fixed.
+//go:generate go get github.com/adriansr/mkwinsyscall
+//go:generate $GOPATH/bin/mkwinsyscall.exe -systemdll -output zsyscall_windows.go syscall_windows.go
+
+//sys PssCaptureSnapshot(processHandle syscall.Handle, captureFlags PSSCaptureFlags, threadContextFlags uint32, snapshotHandle *syscall.Handle) (err error) [failretval!=0] = kernel32.PssCaptureSnapshot
+//sys PssQuerySnapshot(snapshotHandle syscall.Handle, informationClass uint32, buffer *PssThreadInformation, bufferLength uint32) (err error) [failretval!=0] = kernel32.PssQuerySnapshot
```
AndersonQ, line 31: If anyone has a suggestion on how to better organise it, I'd appreciate. I don't quite like how it's now, but also I'm not sure which other organisation could be better.

metric/system/process/process_linux_test.go
```diff
@@ -97,3 +98,38 @@ func TestFetchProcessFromOtherUser(t *testing.T) {
...
+		NumThreads: opt.IntWith(26),
+	}
+
+	got, err := GetInfoForPid(resolve.NewTestResolver("testdata"), 42)
+	require.NoError(t, err, "GetInfoForPid returned an error when it should have succeeded")
+
+	assert.Equal(t, want, got)
+}
+
+func TestParsePIDInfos(t *testing.T) {
```
belimawr, line 118: I don't fully get the name of this test. If I understood it correctly, it's testing `parseProcStat`, however it is called `ParsePIDInfos`.
```suggestion
func TestParseProcStat(t *testing.T) {
```

metric/system/process/process_darwin_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestFetchNumThreads only checks some value is returned for NumThreads. However,
+// it cannot verify the correctness of the value.
+func TestFetchNumThreads(t *testing.T) {
```
cmacknz, line 30: You could test this by writing a program that runs a known number of threads. The trick is that it can't be a Go program to get control over it. A simple C program with a main function that just sleeps until it is terminated is guaranteed to be only one thread, or you could spawn a known number of threads directly with the pthreads API, or you could write it in C++ and use with std:thread which is probably easier to do.

Hardest part would be getting CI to compile the program but this repository already depends on CGO so there must be a C compiler available already.

cmacknz: This would also work for Linux, and if you used `std:thread` in C++ it would also work for Windows. As mentioned, the biggest hurdle will be ensuring you have a C++ toolchain available with the appropriate version of C++ available for `std:thread` support but I hope this is already the case.

metric/system/process/testdata/winbinaries/ProcessSnapshotting.cpp
```diff
@@ -0,0 +1,168 @@
+// from https://github.com/zodiacon/Win10SysProgBookSamples/blob/56883f5126f5e29a03d01896f87ce7b82f51fa10/Chapter20/snapproc/snapproc.cpp
```
cmacknz, line 1: What is the license of this code? Are you allowed to copy it here?

It seems like you might be better off with just linking to the book with instructions on how to build it unless you've modified this significantly.

belimawr (approved):
All good from my side, but I like Craig's idea to have a simple C++ program spawning a known number of threads for testing.

cmacknz:
The images used for CI are defined in https://github.com/elastic/elastic-agent-system-metrics/blob/dc997bdd11e6a1fb3ee32c2d124bd9def96b8d95/.buildkite/pipeline.yml#L6-L7

For Linux at least the Go image contains a pre-installed version of g++ you could use. 

I have no idea what is in that Windows container, but you could modify the powershell script that runs the tests to find out by trying to invoke a few of the toolchains you expect might exist.

https://github.com/elastic/elastic-agent-system-metrics/blob/dc997bdd11e6a1fb3ee32c2d124bd9def96b8d95/.buildkite/scripts/run-win-tests.ps1#L10-L18

We have a dependency manager called chocolately, it's possible you could use it to install mingw.

As for actually compiling the binaries it might be easier to do it with an explicit mage target like `mage buildTestBinaries` similar to the `mage buildSystemTestBinaries` that exists in Beats.

Even if you can't get it to work in CI you can document the commands used to build the executables in mage and just not invoke them in CI (although I hope doing it for Linux at least won't be that bad).

metric/system/process/process_windows_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/elastic/elastic-agent-system-metrics/metric/system/resolve"
+)
+
+func TestGetInfoForPid_numThreads(t *testing.T) {
+	const want = 42
+	// On windows the programs always start with more threads than the code
+	// is launching. It seems to be due to it loading dll/tables in parallel
+	// or de default thread pool. See the following for more info:
+	// https://stackoverflow.com/questions/42789199/why-there-are-three-unexpected-worker-threads-when-a-win32-console-application-s/42789684#42789684
+	expected := 45
```
belimawr, line 36: Can you ensure it's always 45 threads?

One option is to run the program for a while and use the eventually assertion to wait for when the number of threads goes back to 42

metric/system/process/process_linux_darwin_test.go
```diff
@@ -0,0 +1,51 @@
...
+		resolve.NewTestResolver("/"), cmd.Process.Pid)
+	require.NoError(t, err, "failed to GetInfoForPid")
+
+	if !got.NumThreads.Exists() {
+		bs, err := json.Marshal(got)
+		if err != nil {
+			t.Logf("could not marshal ProcState: %v", err)
+		}
+		t.Fatalf("num_thread was not collected. Collected info: %s", bs)
+	}
```
belimawr, line 46: What does `bs` stand for here?

AndersonQ: bytes

metric/system/process/process_windows_test.go
AndersonQ: it takes around 30s on my machine to have the number go back to 42. As it's been pretty deterministic, on my machine and on CI, I'm doing this double test. It accounts exactly for the extra threads, if something changes in the future the test sill fails and there should be enough explanation in the test for anyone else o understand the problem and fix it

PR #103, opened 2023-10-26
reduce resource usage

florianl:
## What does this PR do?

Reduce resource usage by preallocating memory and using caches.

## Why is it important?

`elastic/elastic-agent-system-metrics` is a direct dependency of beats like `metricbeat`.

As shown in the above screenshot, `metricbeat` spends most of its time in `elastic/elastic-agent-system-metrics` doing syscalls to walk directories.

## Checklist

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~
- [ ] ~~I have added an entry in `CHANGELOG.md`~~

### DIFF
--- a/metric/system/cgroup/util.go
+++ b/metric/system/cgroup/util.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/elastic/elastic-agent-libs/logp"
 	"github.com/elastic/elastic-agent-system-metrics/metric/system/resolve"
@@ -276,12 +277,20 @@ the container as /sys/fs/cgroup/unified and start the system module with the hos
 				controllerPath = r.rootfsMountpoint.ResolveHostFS(filepath.Join("/sys/fs/cgroup/unified", path))
 			}
 
+			// Check if there is an entry for controllerPath already cached.
 			if tmp, ok := r.v2ControllerPathCache.Load(controllerPath); ok {
-				pathList, ok := tmp.(PathList)
+				cacheEntry, ok := tmp.(pathListWithTime)
 				if ok {
-					cPaths.V2 = pathList.V2
-					continue
+					// If the cached entry for controllerPath is not older than 5 minutes,
+					// return the cached entry.
+					if time.Since(cacheEntry.added) < time.Duration(5*time.Minute) {
+						cPaths.V2 = cacheEntry.pathList.V2
+						continue
+					}
 				}
+				// Consider the existing entry for controllerPath invalid. The entry can
+				// (1) not be casted to pathListWithTime or (2) is older than 5 minutes.
+				r.v2ControllerPathCache.Delete(controllerPath)
 			}
 
 			cgpaths, err := os.ReadDir(controllerPath)
@@ -296,7 +305,10 @@ the container as /sys/fs/cgroup/unified and start the system module with the hos
 					cPaths.V2[controllerName] = ControllerPath{ControllerPath: path, FullPath: controllerPath, IsV2: true}
 				}
 			}
-			r.v2ControllerPathCache.Store(controllerPath, cPaths)
+			r.v2ControllerPathCache.Store(controllerPath, pathListWithTime{
+				added:    time.Now(),
+				pathList: cPaths,
+			})
 			// cgroup v1
 		} else {
 			subsystems := strings.Split(fields[1], ",")
--- a/metric/system/cgroup/util.go
+++ b/metric/system/cgroup/util.go
@@ -278,20 +278,21 @@ the container as /sys/fs/cgroup/unified and start the system module with the hos
 			}
 
 			// Check if there is an entry for controllerPath already cached.
-			if tmp, ok := r.v2ControllerPathCache.Load(controllerPath); ok {
-				cacheEntry, ok := tmp.(pathListWithTime)
-				if ok {
-					// If the cached entry for controllerPath is not older than 5 minutes,
-					// return the cached entry.
-					if time.Since(cacheEntry.added) < time.Duration(5*time.Minute) {
-						cPaths.V2 = cacheEntry.pathList.V2
-						continue
-					}
+			r.v2ControllerPathCache.Lock()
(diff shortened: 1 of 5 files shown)


florianl:
The impact of this proposed change is shown in the following screenshot:



The "regular" `metricbeat` (left side) represents the current status that is shipped with https://snapshots.elastic.co/8.12.0-f9668451/summary-8.12.0-SNAPSHOT.html. On the right sight of the screenshot the proposed change with me `metricbeat-own` - notice here the less purple `vmlinux` frames. This translates directly to less syscalls and therefore less CPU usage of `metricbeat-own`.

florianl:
Differential function view of metricbeat and metricbeat-own:

metric/system/cgroup/util.go
```diff
@@ -276,6 +277,14 @@ the container as /sys/fs/cgroup/unified and start the system module with the hos
 				controllerPath = r.rootfsMountpoint.ResolveHostFS(filepath.Join("/sys/fs/cgroup/unified", path))
 			}
 
+			if tmp, ok := r.v2ControllerPathCache.Load(controllerPath); ok {
```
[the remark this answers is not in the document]
florianl: Not sure I can follow the question :thinking: 
cgroup v2 can change. And regardless of this PR or the existing approach, data will be missed or incorrect. It will be missed or incorrect, as walking `controllerPath` to get all the `stat` files and reading the content of these files is happening independently. There is no change in this behaviour.
To address the concern that entries could stay forever in `r.v2ControllerPathCache` we could introduce a TTL. Would this be ok for you? And if so, which TTL would you consider? 5 minutes, 30 minutes, 1 hours, 2 hours or even 1 day?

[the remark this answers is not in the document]
florianl: @kruskall with https://github.com/elastic/elastic-agent-system-metrics/pull/103/commits/0cb8e6e0ab759ce558b522a029900057be7c7aac I have added a timestamp to each cache entry to make sure entries do not stay forever in the cache. Does this address your concern?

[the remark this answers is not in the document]
florianl: > My concern was about the system reusing PIDs causing the files in the cache to be inaccurate.

The porposed change does not have an effect on reused PIDs. At the moment of scanning all PIDs, eg. via `FetchPids()`, every PID might have some relation to a cgroup. And for cgroup v2 it is `controllerPath` that describes the dedicated cgroup v2. Usually multiple PIDs share the same cgroup v2. This proposed change reduces CPU usage load  by introducing a cache, to not look up every time `controllerPath` but cache its results (for 5 minutes).

fearful-symmetry:
caching the paths seems reasonable, I think? Particularly for V2, when we're just reading everything out of the same unified hierarchy.

belimawr (approved):

metric/system/cgroup/util.go
```diff
@@ -275,6 +277,22 @@ the container as /sys/fs/cgroup/unified and start the system module with the hos
...
+					// If the cached entry for controllerPath is not older than 5 minutes,
+					// return the cached entry.
+					if time.Since(cacheEntry.added) < time.Duration(5*time.Minute) {
+						cPaths.V2 = cacheEntry.pathList.V2
+						continue
+					}
+				}
+				// Consider the existing entry for controllerPath invalid. The entry can
+				// (1) not be casted to pathListWithTime or (2) is older than 5 minutes.
+				r.v2ControllerPathCache.Delete(controllerPath)
```
ycombinator, line 293: Instead of using a `sync.Map` why not use a struct wrapping regular map and a mutex? This would give us better type safety and we wouldn't need the type assertion in this code here. It should also make the logic simpler here, as it would be just about whether the cache entry has expired or not.

metric/system/cgroup/util.go
```diff
@@ -275,6 +277,22 @@ the container as /sys/fs/cgroup/unified and start the system module with the hos
 				controllerPath = r.rootfsMountpoint.ResolveHostFS(filepath.Join("/sys/fs/cgroup/unified", path))
 			}
 
+			// Check if there is an entry for controllerPath already cached.
+			if tmp, ok := r.v2ControllerPathCache.Load(controllerPath); ok {
+				cacheEntry, ok := tmp.(pathListWithTime)
+				if ok {
+					// If the cached entry for controllerPath is not older than 5 minutes,
+					// return the cached entry.
+					if time.Since(cacheEntry.added) < time.Duration(5*time.Minute) {
```
ycombinator, line 286: Nit: the `time.Duration` explicit cast here is redundant.
```suggestion
					if time.Since(cacheEntry.added) < 5*time.Minute {
```

metric/system/cgroup/util.go
florianl: I did introduce the synchronization mechanism  as it is not guaranteed that `Reader` is only used in a sequential order. And as multiple Go routine can use `Reader` and a regular `map` is not Go thread safe, I went with `sync.Map`. I did choose `sync.Map` over a struct with a regular map and a mutex, as it is harder to abuse and therefore more safe. While I agree on the added complexity around `sync.Map` handling, it prevents direct accessing a map in a struct, without holding the sync mutex.
Is this ok for you?

ycombinator: Totally get the need for a synchronization here!  It's just that the documentation of `sync.Map` itself doesn't recommend using it over a regular map with a mutex, citing type safety as one of the reasons :)

https://pkg.go.dev/sync#Map

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.

I also searched through the existing `elastic-agent-system-metrics` and `elastic-agent` codebases and I'm not seeing any uses of `sync.Map`, probably for this same reason.  So I'm hesitant to break that pattern here.

florianl: While `sync.Map` is not used in the two named repositories, other beats and Elastic projects are using it - https://github.com/search?type=code&q=org%3Aelastic+lang%3Ago+sync.Map

I updated the PR to keep the consistency of `elastic-agent-system-metrics`

PR #104, opened 2023-11-02
Fix metrics collection for privileged process on Windows

belimawr:
## What does this PR do?

It improves metric collection on Windows hosts so we can collect some metrics from privileged process like Elastic Endpoint or Elastic-Agent. In order to achieve that metrics collection for Windows are grouped by the access level required to collect metrics. First the metrics that can be collected with `PROCESS_QUERY_LIMITED_INFORMATION` are collected, then the others are collected. In case the second batch fails, we still report the partial metrics.

## Why is it important?

It allows us to collect CPU and memory metrics from Endpoint and Elastic-Agent, which improves our monitoring dashboards.

## Checklist

## Author's Checklist
- [x] Make sure nothing broke on:
  - [x] Linux
  - [x] MacOS

## Related issues

- Closes https://github.com/elastic/beats/issues/17314

### DIFF
--- a/metric/system/process/process.go
+++ b/metric/system/process/process.go
@@ -169,10 +169,11 @@ func (procStats *Stats) GetSelf() (ProcState, error) {
 func (procStats *Stats) pidIter(pid int, procMap ProcsMap, proclist []ProcState) (ProcsMap, []ProcState) {
 	status, saved, err := procStats.pidFill(pid, true)
 	if err != nil {
-		if !errors.Is(err, NotEnoughPrivilegesErr{}) {
+		if !errors.Is(err, NonFatalErr{}) {
 			procStats.logger.Debugf("Error fetching PID info for %d, skipping: %s", pid, err)
 			return procMap, proclist
 		}
+		procStats.logger.Debugf("Non fatal error fetching PID some info for %d, metrics are valid, but partial: %s", pid, err)
 	}
 	if !saved {
 		procStats.logger.Debugf("Process name does not match the provided regex; PID=%d; name=%s", pid, status.Name)
@@ -184,21 +185,25 @@ func (procStats *Stats) pidIter(pid int, procMap ProcsMap, proclist []ProcState)
 	return procMap, proclist
 }
 
-// NotEnoughPrivilegesErr is returned when the current access
-// rights are not enough to get some metrics.
-// This error can be safely ignored and only means the function/method
-// could not gather all metrics, however whatever has been gethered
-// is still valid.
-type NotEnoughPrivilegesErr struct {
+// NonFatalErr is returned when there was an error
+// collecting metrics, however the metrics already
+// gathered and returned are still valid.
+// This error can be safely ignored, this will result
+// in having partial metrics for a process rather than
+// no metrics at all.
+//
+// It was introduced to allow for partial metrics collection
+// on privileged process on Windows.
+type NonFatalErr struct {
 	Err error
 }
 
-func (c NotEnoughPrivilegesErr) Error() string {
+func (c NonFatalErr) Error() string {
 	return "Not enough privileges to fetch information: " + c.Err.Error()
 }
 
-func (c NotEnoughPrivilegesErr) Is(other error) bool {
-	_, is := other.(NotEnoughPrivilegesErr)
+func (c NonFatalErr) Is(other error) bool {
+	_, is := other.(NonFatalErr)
 	return is
 }
 
--- a/metric/system/process/process.go
+++ b/metric/system/process/process.go
@@ -167,10 +167,9 @@ func (procStats *Stats) GetSelf() (ProcState, error) {
 // pidIter wraps a few lines of generic code that all OS-specific FetchPids() functions must call.
 // this also handles the process of adding to the maps/lists in order to limit the code duplication in all the OS implementations
 func (procStats *Stats) pidIter(pid int, procMap ProcsMap, proclist []ProcState) (ProcsMap, []ProcState) {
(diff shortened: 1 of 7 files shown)


metric/system/process/process_windows.go
```diff
@@ -140,11 +134,25 @@ func FillPidMetrics(_ resolve.Resolver, pid int, state ProcState, _ func(string)
 
 	state.CPU.StartTime = unixTimeMsToTime(startTime)
 
+	return state, nil
+}
+
+// FillOtherMetricsMoreAccess
+// All calls that need more access rights than
+// windows.PROCESS_QUERY_LIMITED_INFORMATION
+func FillOtherMetricsMoreAccess(pid int, state ProcState) (ProcState, error) {
```
belimawr, line 143: This name needs some work. Tomorrow I'll push a new commit with a better name.

belimawr:
It's ready for review.

metric/system/process/process.go
```diff
@@ -166,10 +167,13 @@ func (procStats *Stats) GetSelf() (ProcState, error) {
 // pidIter wraps a few lines of generic code that all OS-specific FetchPids() functions must call.
 // this also handles the process of adding to the maps/lists in order to limit the code duplication in all the OS implementations
 func (procStats *Stats) pidIter(pid int, procMap ProcsMap, proclist []ProcState) (ProcsMap, []ProcState) {
+	// That's the method that fails, however it seems to return a partial state
```
faec, line 170: This looks like maybe a leftover or debugging comment? If not could it be clarified

metric/system/process/process.go
```diff
@@ -181,6 +185,24 @@ func (procStats *Stats) pidIter(pid int, procMap ProcsMap, proclist []ProcState)
 	return procMap, proclist
 }
 
+// NotEnoughPrivilegiesErr is returned when the current access
```
faec, line 188: Spelling: "Privileges" (here and elsewhere)

metric/system/process/process.go
```diff
@@ -189,6 +211,7 @@ func (procStats *Stats) pidFill(pid int, filter bool) (ProcState, bool, error) {
 	// Fetch proc state so we can get the name for filtering based on user's filter.
 
 	// OS-specific entrypoint, get basic info so we can at least run matchProcess
+	// last non-os-specific call. Here "the problem begins"
```
faec, line 214: Could this be clarified? It's not clear what problem this means, especially for someone just looking at the code and not this specific PR

metric/system/process/process_windows.go
```diff
@@ -201,8 +209,7 @@ func getProcTimes(pid int) (uint64, uint64, uint64, error) {
 
 func procMem(pid int) (uint64, uint64, error) {
 	handle, err := syscall.OpenProcess(
-		windows.PROCESS_QUERY_LIMITED_INFORMATION|
-			windows.PROCESS_VM_READ,
```
faec, line 205: is `PROCESS_VM_READ` no longer needed anywhere? I don't see it added to a more privileged function call anywhere

metric/system/process/process.go
belimawr: another debug comment I forgot to delete :man_facepalming:

metric/system/process/process_windows.go
belimawr: `PROCESS_VM_READ` is only needed for Windows Server 2003 and Windows XP, we support neither of them.

>A handle to the process. The handle must have the PROCESS_QUERY_INFORMATION or PROCESS_QUERY_LIMITED_INFORMATION access right. For more information, see [Process Security and Access Rights](https://learn.microsoft.com/en-us/windows/desktop/ProcThread/process-security-and-access-rights).
>
>Windows Server 2003 and Windows XP: The handle must have the PROCESS_QUERY_INFORMATION and PROCESS_VM_READ access rights.

https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getprocessmemoryinfo

belimawr: I added a comment explaining it

belimawr:
@faec I addressed all of your comments on 9ef7b7172

metric/system/process/process_windows.go
```diff
@@ -140,11 +134,25 @@ func FillPidMetrics(_ resolve.Resolver, pid int, state ProcState, _ func(string)
...
+}
+
+// FillMetricsRequiringMoreAccess
+// All calls that need more access rights than
+// windows.PROCESS_QUERY_LIMITED_INFORMATION
+func FillMetricsRequiringMoreAccess(pid int, state ProcState) (ProcState, error) {
 	argList, err := getProcArgs(pid)
 	if err != nil {
-		return state, fmt.Errorf("error fetching process args: %w", err)
+		return state, fmt.Errorf("error fetching process args: %w", NotEnoughPrivilegesErr{Err: err})
```
leehinman, line 146: If I'm reading this right, it turns all errors from `getProcArgs` into not enough privileges errors.  But if I look at `getProcArgs` it might fail for reasons other than permissions.  Can we do this for cases where it is really just a permission problem? or maybe change the name?

belimawr: You're right. On my first implementation I had something like `CanIgnoreErr` to transmit the idea that even though we got an error we can still trust and use the metrics that were returned.

I'll try to came up with a better name.

belimawr: I fixed it on 15e3d6200 and also improved the logging.

leehinman (approved):
LGTM

leehinman (approved):

PR #107, opened 2023-11-06
Fix CmdLine generation and caching

belimawr:
## What does this PR do?

When collecting process metrics the generation of `cmdLine` field and its caching were happening in the wrong place. This commit fixes it.

## Why is it important?

It fixes an issue with metrics collection

## Checklist

~~## Author's Checklist~~

## Related issues
- Relates https://github.com/elastic/beats/issues/17314


fearful-symmetry (approved):

metric/system/process/process.go
```diff
@@ -222,7 +222,6 @@ func (procStats *Stats) pidFill(pid int, filter bool) (ProcState, bool, error) {
 	if procStats.skipExtended {
 		return status, true, nil
 	}
-	status = procStats.cacheCmdLine(status)
```
belimawr, line 225: I'm wondering if that should have stayed here or somewhere after setting `status.Cmdline`. I'll look into it and update this thread.

belimawr: Fixes by moving it back where it was and adding some comments.

PR #114, opened 2023-11-16
Add metrics from /proc/[pid]/io to system/process metrics on linux

fearful-symmetry:
## What does this PR do?

Part of https://github.com/elastic/beats/issues/7461

This ended up being fairly simple; we just fetch per-process IO from procfs, and these values appear to be largely identical to what's reported by the netlink taskstat.

This just reads metrics from `/proc/[pid]/io` and reports them as part of other process metrics, same as how things like memory usage are reported.

## Why is it important?

We want per-process I/O metrics.

## Checklist


metric/system/process/process_types.go
```diff
@@ -79,6 +80,24 @@ type CPUTotal struct {
...
+	//ReadSyscalls counts the number of read operations
+	ReadSyscalls opt.Uint `struct:"read_ops,omitempty"`
+	//WriteSyscalls counts the number of write operations
+	WriteSyscalls opt.Uint `struct:"write_ops,omitempty"`
+	// ReadBytes is the count of bytes that were actually fetched from the storage layer
+	ReadBytes opt.Uint `struct:"read_bytes,omitempty"`
+	// WriteBytes is the count of bytes that were actually written to the storage layer
+	WriteBytes opt.Uint `struct:"write_bytes,omitempty"`
+	// the number of bytes which this process caused to not happen, by truncating pagecache
+	CancelledWriteBites opt.Uint `struct:"cancelled_write_bytes,omitempty"`
```
rdner, line 98:
```suggestion
	// CancelledWriteBytes is the number of bytes which this process caused to not happen, by truncating pagecache
	CancelledWriteBytes opt.Uint `struct:"cancelled_write_bytes,omitempty"`
```

Is it `Bites` on purpose? If accepted this change requires renaming in the rest of the code.

rdner (approved):

PR #116, opened 2023-11-20
metric/system: preallocate arrays

florianl:
## What does this PR do?

Whenever possible, preallocate array to the correct size instead of letting arrays "grow". This reduces the load on the Garbage Collector and improves memory usage efficiency.

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~
- [ ] ~~I have added an entry in `CHANGELOG.md`~~

Issue #109, opened 2023-11-16
broken on darwin & windows


leehinman:
If we try to use elastic-agent-system-metrics for darwin or windows we get the following errors:

process_windows.go:41:32: undefined: names
process_windows.go:42:36: undefined: names

or

process_darwin.go:67:32: undefined: names
process_darwin.go:68:36: undefined: names

This was introduced by https://github.com/elastic/elastic-agent-system-metrics/pull/103

leehinman:
@florianl can you please look at this since it was introduced by your PR?

belimawr:
Revert PR to unblock `main`/new PRs: https://github.com/elastic/elastic-agent-system-metrics/pull/113

florianl:
Pushed a PR with a fix in https://github.com/elastic/elastic-agent-system-metrics/pull/116.

florianl:
Revert of the revert with https://github.com/elastic/elastic-agent-system-metrics/pull/117

florianl:
With https://github.com/elastic/elastic-agent-system-metrics/pull/116 (the fix) and https://github.com/elastic/elastic-agent-system-metrics/pull/117 (the reintroduction of the revert) merged, I'm closing this issue.

PR #117, opened 2023-11-24
metric/system/cgroup: introduce cache for cgroup v2

florianl:
## What does this PR do?

Reintroduce the improvements from https://github.com/elastic/elastic-agent-system-metrics/pull/103. This PR got reverted with https://github.com/elastic/elastic-agent-system-metrics/pull/113 because of https://github.com/elastic/elastic-agent-system-metrics/issues/109. The issue, that got reported in https://github.com/elastic/elastic-agent-system-metrics/issues/109, was fixed with https://github.com/elastic/elastic-agent-system-metrics/pull/116.

So its time to bring back the performance improvements by reintroducing the cache for cgroup v2.

## Checklist

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~
- [ ] ~~I have added an entry in `CHANGELOG.md`~~

### DIFF
--- a/metric/system/cgroup/util.go
+++ b/metric/system/cgroup/util.go
@@ -230,7 +230,7 @@ func SubsystemMountpoints(rootfs resolve.Resolver, subsystems map[string]struct{
 
 // ProcessCgroupPaths returns the cgroups to which a process belongs and the
 // pathname of the cgroup relative to the mountpoint of the subsystem.
-func (r Reader) ProcessCgroupPaths(pid int) (PathList, error) {
+func (r *Reader) ProcessCgroupPaths(pid int) (PathList, error) {
 	cgroupPath := filepath.Join("proc", strconv.Itoa(pid), "cgroup")
 	cgroup, err := os.Open(r.rootfsMountpoint.ResolveHostFS(cgroupPath))
 	if err != nil {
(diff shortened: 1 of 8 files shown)


metric/system/cgroup/util.go
```diff
@@ -275,6 +276,23 @@ the container as /sys/fs/cgroup/unified and start the system module with the hos
...
+				// return the cached entry.
+				if time.Since(cacheEntry.added) < 5*time.Minute {
+					cPaths.V2 = cacheEntry.pathList.V2
+					r.v2ControllerPathCache.Unlock()
+					continue
+				}
+			}
+			// Consider the existing entry for controllerPath invalid, as it is
+			// older than 5 minutes.
+			delete(r.v2ControllerPathCache.cache, controllerPath)
```
leehinman, line 293: Nit.  Even though the delete will be a no-op if ok is false (not in cache) it reads oddly.  If the entry isn't in a map I don't expect to try to immediately delete the entry.  The delete is really there if we don't hit the `continue`.  What do you think about moving the delete inside the `if ok` block?  That way the Unlock is right after the `if ok` and the delete is closer to the condition that causes it to happen.

leehinman (approved):
Approved.  Optional suggestion.

florianl:
sorry - but making linter happy seems to be impossible ...

cmacknz:
If the linter is the only problem we can force merge this, at least for me the linter is not even a blocking step (the squash and merge button is available) so there is no force merge required.

PR #120, opened 2023-11-28
Fix process metrics collection when both cgroup V1 and V2 controllers exist

ydubreuil:
## What does this PR do?

This PR fixes a bug in the cgroup module of process metrics collection when both cgroup V1 and V2 controllers exist but `/sys/fs/cgroups/unified` is not mounted.

This case can exist on Kubernetes nodes where systemd is configured to use the cgroup V1 hierarchy but a service needs to mount the cgroup V2 hierarchy for its own use (This is what Cilium does for example). In that case, the `proc` tree contains a reference to a empty cgroup V2 controller which confuses metricbeat which report a `PID XXX contains a cgroups V2 path (0::/) but no V2 mountpoint was found.\nThis may be because metricbeat is running inside a container on a hybrid system.\nTo monitor cgroups V2 processess in this way, mount the unified (V2) hierarchy inside\nthe container as /sys/fs/cgroup/unified and start the system module with the hostfs setting.` error.

## Why is it important?

This fixes gathering process metrics on Amazon Kubernetes nodes running Cilium.

## Checklist

## Author's Checklist

This PR is split in 2 commits:

* a test case to reproduce the issue with the existing code
* a logic change in to fix the issue

Running tests with only the first commit leads to this error:

```
?   	github.com/elastic/elastic-agent-system-metrics/metric/system/cpu	[no test files]
--- FAIL: TestReaderGetStatsV1MalformedHybrid (0.00s)
    reader_test.go:92: 
        	Error Trace:	reader_test.go:92
        	Error:      	Received unexpected error:
        	            	error fetching cgroupV2 controllers for cgroup location '' and path line '0::/': open testdata/amzn2/sys/fs/cgroup/unified: no such file or directory
        	Test:       	TestReaderGetStatsV1MalformedHybrid
        	Messages:   	error in GetV1StatsForProcess
--- FAIL: TestProcessCgroupHybridPaths (0.00s)
    util_test.go:212: error in ProcessCgroupPaths: error fetching cgroupV2 controllers for cgroup location '' and path line '0::/': open testdata/amzn2/sys/fs/cgroup/unified: no such file or directory
FAIL

```

## Related issues

Relates to https://github.com/elastic/beats/issues/37211


metric/system/cgroup/util.go
```diff
@@ -229,7 +229,7 @@ func SubsystemMountpoints(rootfs resolve.Resolver, subsystems map[string]struct{
 
 // ProcessCgroupPaths returns the cgroups to which a process belongs and the
 // pathname of the cgroup relative to the mountpoint of the subsystem.
-func (r Reader) ProcessCgroupPaths(pid int) (PathList, error) {
+func (r Reader) ProcessCgroupPaths(version CgroupsVersion, pid int) (PathList, error) {
```
fearful-symmetry, line 232: A little ambivalent about how we're doing this. Right now, this function is basically version agnostic, and will return a list of both V1 and V2 cgroups, which are part of `PathList`. This might not matter for `GetV1StatsForProcess` and `GetV2StatsForProcess`, but `ProcessCgroupPaths` is called by the `add_docker_metadata` and `add_process_metadata` in beats, and I'm a little worried about us somehow changing the behavior of the processors under hybrid systems if we revert to either _only_ V1 or V2

My suggestion is that we create two functions (they can probably just be wrappers for the same private function or something) that can either version-agnostic, and return the full `PathList`, or that are version-specific, and just return the a map of the given version paths.

fearful-symmetry (changes requested):

metric/system/cgroup/util.go
ydubreuil: Yeah, I wasn't sure if `ProcessCgroupPaths` was part of the public API for the module or not. As it's called  from beats, I'll rework the PR to not change the signature.

ydubreuil: I changed the approach in https://github.com/elastic/elastic-agent-system-metrics/pull/120/commits/ce87b3c8acb60c5f7d6a774f539a41b800804d0c to contain the changes in `ProcessCgroupPaths`. I had to call `CgroupsVersion` to know if we were in cgroup V1 or V2 mode. I could have refactored  `ProcessCgroupPaths` instead and collected all the cgroup controller before resolving paths, but I didn't want to copy the logic from `CgroupsVersion`

fearful-symmetry (approved):
This seems cautious enough, I'd say.

PR #125, opened 2024-01-03
[updatecli] Bump Golang version to 1.21.5

cmacknz:
# [updatecli] Bump Golang version to 1.21.5

### What
Bump go release version with the latest release.

### Further details
See [changelog](https://github.com/golang/go/issues?q=milestone%3AGo1.21.5+label%3ACherryPickApproved) for 1.21.5
Generated automatically with manual CLI

## Report

	Source:
		✔ [gomod] Get version in go.mod format(shell)
		✔ [latestGoVersion] Get Latest Go Release(githubrelease)
		✔ [minor] Get minor version in .go-version(shell)

	Condition:
		✔ [dockerTag] Is docker image golang:1.21.5 published(dockerimage)
		✔ [goDefaultVersion-check] Check if defined golang version differs(shell)

	Target:
		⚠ [update-go-version] Update .go-version(file)
		⚠ [update-golang.ci] Update .golangci.yml(file)
		⚠ [update-gomod] Update go.mod(file)

## Changelog

<details><summary>Click to expand</summary>

````

no GitHub Release found for go1.21.5 on "https://github.com/golang/go"

````

</details>

## Remark

This pull request was automatically created using [Updatecli](https://www.updatecli.io).

Please report any issues with this tool [here](https://github.com/updatecli/updatecli/issues/)

PR #126, opened 2024-01-05
Fix process filter test

fearful-symmetry:
## What does this PR do?

Fixes an issue that came to light with https://github.com/elastic/elastic-agent-system-metrics/pull/125

In certain instances, `FilterTest` will fail, if the process with the most memory usage and most CPU usage are the same.

### DIFF
--- a/metric/system/process/process_test.go
+++ b/metric/system/process/process_test.go
@@ -657,8 +657,7 @@ func initTestResolver() (Stats, error) {
 	return testConfig, err
 }
 
-// nolint: deadcode,structcheck,unused // needed by other platforms
-func sliceContains(s []string, e string) bool {
+func sliceContains(s []string, e string) bool { //nolint: deadcode,structcheck,unused // needed by other platforms
 	for _, v := range s {
 		if e == v {
 			return true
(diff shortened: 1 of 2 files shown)


cmacknz (approved):

metric/system/process/process_test.go
```diff
@@ -357,6 +360,10 @@ func TestProcCpuPercentage(t *testing.T) {
 }
 
 func TestIncludeTopProcesses(t *testing.T) {
+	// noop line so `runThreads` is not marked as unsued for CI on linux
+	// can't use linter directives, since they'll fail on platforms where `runthreads` IS used
```
cmacknz, line 364: You should be able to turn off the lint warning for unused directives.

cmacknz: https://github.com/elastic/elastic-agent-system-metrics/blob/085e4529f3c4f91dd377cadbbe7a2bf321989438/.golangci.yml#L102-L104

fearful-symmetry: ah! found the config option, yah

Issue #132, opened 2024-03-08
cgroups not accounting for relative paths that happen when reading from a container mount


fearful-symmetry:
The `/proc/[pid]/cgroup` file contains the paths to the cgroups used by the process. Most importantly, this path is relative to the cgroup of the process checking the cgroup. From the [man page](https://man7.org/linux/man-pages/man7/cgroups.7.html):
```
              [3]  This field contains the pathname of the control group
                   in the hierarchy to which the process belongs.  This
                   pathname is relative to the mount point of the
                   hierarchy.
```

This means that if you mount in a process from the host system into a container, you get a relative path:

```
 docker exec -it 5b938d5468d3 /bin/bash                                                                                                                                                                                           
metricbeat@motmot:~$ cat /hostfs/proc/114856/cgroup 
0::/../elastic-agent.service
```

Right now this library doesn't seem to cope with this, as it assumes a universal base path:
```
{"log.level":"debug","@timestamp":"2024-03-08T20:32:58.467Z","log.logger":"processes","log.origin":{"file.name":"process/process.go","file.line":173},"message":"Error fetching PID info for 1023161, skipping: cgroups.GetStatsForPid: error fetching cgroupV2 controllers for cgroup location '/hostfs/sys/fs/cgroup' and path line '0::/../../user.slice/user-1000.slice/session-212.scope': open /hostfs/sys/user.slice/user-1000.slice/session-212.scope: no such file or directory","service.name":"metricbeat","ecs.version":"1.6.0"}
```

I'm not sure why we didn't catch this before? Perhaps there's a conflux of different docker versions, [bugs](https://github.com/moby/moby/issues/34584) and configs that have escaped most people.

fearful-symmetry:
A temporary workaround is to set `--cgroupns host`, but that obviously has some downsides.

PR #134, opened 2024-03-18
Make cgroup errors skippable in system/process

fearful-symmetry:
## What does this PR do?

This makes it so cgroup-specific errors don't fail an entire PID, so the PID will still be reported even if we have cgroup errors.

## Why is it important?

This is a sort of band-aid for https://github.com/elastic/elastic-agent-system-metrics/issues/132, as I'm not sure how long a "proper" fix for that will take, as it might require significant changes to the cgroup client, as well as a considerable degree of testing.

## Checklist


metric/system/process/process_linux_test.go
```diff
@@ -234,6 +234,25 @@ func TestParseProcStat(t *testing.T) {
 	assert.Equal(t, want, got, "")
 }
 
+func TestCgroupsBadCgroupsConfig(t *testing.T) {
+	_ = logp.DevelopmentSetup()
```
leehinman, line 238: optional, but since you are setting up logp, is it worthwhile to do `DevelopmentSetup(ToObserverOutput())` and then use `ObserverLogs().TakeAll()` to get the logs and make sure the error message you expect is in the log output?

leehinman (approved):
optional testing improvement

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - Fix CmdLine generation and caching for system.process
  - Fix process metrics collection when both cgroup V1 and V2 controllers exist
  - Skip permissions errors when reading /proc/pid/io
+ - Skip errors in cgroup data collection #134
```
leehinman, line 24:
```suggestion
 - Skip non-fatal errors in cgroup data collection #134
```

leehinman (approved):
code LGTM, one little suggestion on changelog

Issue #139, opened 2024-04-10
`SubsystemMountpoints()` can return wong V2 path for non-root container environments


fearful-symmetry:
Right now, `SubsystemMountpoints()` will calculate the V2 mountpoint by iterating through `/proc/self/mountpoints` and and updating the `V2Loc` as it finds it in the file, ending with the last result. Problem is, some `mountpoints` files look like this:

```
1682 1681 0:26 / /sys/fs/cgroup ro,nosuid,nodev,noexec,relatime - cgroup2 cgroup rw,seclabel
1718 1686 0:26 / /hostfs/sys/fs/cgroup rw,nosuid,nodev,noexec,relatime master:4 - cgroup2 cgroup2 rw,seclabel
1771 1770 0:26 / /hostfs/var/lib/docker/overlay2/1b570230fa3ec3679e354b0c219757c739f91d774ebc02174106488606549da0/merged/sys/fs/cgroup ro,nosuid,nodev,noexec,relatime - cgroup2 cgroup rw,seclabel
```

In this case, the last mountpoint is `/hostfs/var/lib/docker/overlay2/1b570230fa3ec3679e354b0c219757c739f91d774ebc02174106488606549da0/merged/sys/fs/cgroup`

Problem is, this is a docker overlayfs path, and if the container is running as non-root `--user=whomever`, then we can't read from this file.

The hostfs path works fine: 
```
cat /hostfs/sys/fs/cgroup/user.slice/user-1000.slice/session-1763.scope/io.pressure 
some avg10=0.00 avg60=0.00 avg300=0.00 total=2256657
full avg10=0.00 avg60=0.00 avg300=0.00 total=1667639
```

Issue #135, opened 2024-03-20
error checks when reading `/proc/[pid]/exe` don't account for kernel processes or docker permissions shenanigans 


fearful-symmetry:
Right now, under Linux we check to see if we have a permission denied error when we resolve the symlink for `exe`:
https://github.com/elastic/elastic-agent-system-metrics/blob/8a02eaa6c02e7972cc8d0adc90c1c13db836e273/metric/system/process/process_linux_common.go#L209

We do this so we can skip any errors related to permissions issues. However, this doesn't work under two cases:

1. For kernel processes, the `exe` symlink is non-existent
2. Under docker, we'll get either `Permission deined` or `file not found` depending on the permissions docker is running with.

The permissions check in `process_linux_common.go` should check for _both_ permission errors, and file not found errors.

pierrehilbert:
Blocked as required https://github.com/elastic/beats/issues/38241 first

PR #140, opened 2024-04-11
Add a test framework for monitoring the host system from inside a container

fearful-symmetry:
## What does this PR do?
Closes https://github.com/elastic/beats/issues/38241

This adds a lightweight test framework that runs a set of system tests under a container with the goal of monitoring the host system. The goal with these tests is to catch the numerous edge cases that happen when the system metrics function from a `/hostfs` path inside a container.

The tests have a fairly large matrix of configurations, as we need to test both a wide variety of container permission settings, as well as differences in how linux distros will configure cgroups.

The framework here was designed with the goal of being relatively idiomatic; you can just run the framework with `go test` as you would normally.

You can run the tests yourself with `go test -v ./tests`

As you may have noticed, there's a non-zero amount of TODO statements here, since these tests were built to aggravate a bunch of existing bugs, so certain parts of the tests will remain un-implemented until those bugs are fixed.

## Why is it important?

See https://github.com/elastic/beats/issues/38241, we really need test for this particular case.

## List of bugs that are responsible for TODO statements in the tests:

- https://github.com/elastic/elastic-agent-system-metrics/issues/141
- https://github.com/elastic/elastic-agent-system-metrics/issues/135
- https://github.com/elastic/elastic-agent-system-metrics/issues/139
- https://github.com/elastic/elastic-agent-system-metrics/issues/132
- https://github.com/elastic/go-sysinfo/issues/12

## Checklist


dev-tools/systemtests/container.go
```diff
@@ -0,0 +1,303 @@
...
+	// skip if the process has already been created
+	if tr.CreateHostProcess != nil && tr.CreateHostProcess.Process == nil {
+		// We don't need to do this in a channel, but it prevents races between this goroutine
+		// and the rest of test framework
+		startPid := make(chan int)
+		log.Infof("Creating test Process...")
+		go func() {
+			err := tr.CreateHostProcess.Start()
+			// if the process fails to start up, the resulting tests will fail, so just log it
+			require.NoError(tr.Runner, err, "error starting monitor process")
```
belimawr, line 292: `require` uses `t.FailNow` which should not be called from a different goroutine than the one running the test, from the [docs](https://pkg.go.dev/testing#T.FailNow):
> (...) FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines.

It's better to use `assert` here
```suggestion
			assert.NoError(tr.Runner, err, "error starting monitor process")
```

metric/system/process/process.go
```diff
@@ -280,6 +280,7 @@ func (procStats *Stats) pidFill(pid int, filter bool) (ProcState, bool, error) {
 
 	// network data
 	if procStats.EnableNetwork {
+		// TODO: this can't handle hostfs settings
```
belimawr, line 283: Does this TODO still need to be done? Or can this comment be removed?

fearful-symmetry: ...I need to file a bug for it, thanks for reminding me

.buildkite/scripts/system-container-tests.sh
```diff
@@ -0,0 +1,16 @@
...
+source .buildkite/scripts/common.sh
+
+install_go_dependencies
+
+# install docker
+
+# DOCKER_VERSION="25.0"
+
+# curl -fsSL https://get.docker.com -o get-docker.sh
+# sudo sh ./get-docker.sh --version $DOCKER_VERSION
```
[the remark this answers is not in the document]
fearful-symmetry: Depending on how creative I get with the buildkite images later, I might need it, so I'm leaving it in

faec (approved):
Looks good!

leehinman (approved):

PR #146, opened 2024-04-23
Update some dependencies, change how go is installed in windows

michel-laterman:
Update a bunch of dependencies
Use gvm through the powershell on windows to install go.


cmacknz (approved):

pierrehilbert:
We have some failures here:
```
C:\Users\buildkite\go\pkg\mod\go.opentelemetry.io\otel@v1.25.0\attribute\set.go:10:2: package slices is not in GOROOT (C:\Program Files\Go\src\slices)
note: imported by a module that requires go 1.21

**Comment by michel-laterman** [comment]
I think there's an issue with how the windows tests are setup, the incorect version of go is being used:
```
Chocolatey installed 1/1 packages.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
Refreshing environment variables from the registry for powershell.exe. Please wait...
Finished
go version go1.20.5 windows/amd64
```

File: .buildkite/pipeline.yml [hunk 1]
```diff
@@ -2,7 +2,7 @@
 
 env:
   SETUP_GVM_VERSION: 'v0.5.0'
-  GO_VERSION_CHOCO: "1.20.5"
+  GO_VERSION_CHOCO: "1.21.9"
```
michel-laterman, line 5: It looks like chocolaty is inconsistent availability for go's previouus minor releases; 1.21.9 is not available.

If we remove the go version specification for the windows tests the latest package would be used (in this case a 1.22.x release)
@cmacknz @fearful-symmetry what do you think?

leehinman: Chocolaty doesn't package every go version, so even if this works this time we could run into this again.  Can we please consider moving away from chocolaty and to something that will work for every go version?  For example https://github.com/andrewkroh/gvm has been used in the past on Windows and it works fine.

fearful-symmetry: Yeah, using `gvm` or something might be a good idea.

.buildkite/scripts/run-win-tests.ps1
```diff
@@ -30,7 +29,7 @@ function installGoDependencies {
 }
 
 fixCRLF
-withGolang $env:GO_VERSION_CHOCO
+withGolang $env:GO_VERSION
```
michel-laterman, line 32: Good news, the gvm install works!
```
-- Install golang --
--
  | go version go1.21.9 windows/amd64
```

Bad news, the tests (silently) fail:
```

gotestsum : The term 'gotestsum' is not recognized as the name of a cmdlet, function, script file, or operable
--
  | program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
  | At C:\buildkite-agent\builds\bk-agent-prod-gcp-1714061473631994270\elastic\elastic-agent-system-metrics\.buildkite\scri
  | pts\run-win-tests.ps1:37 char:1
  | + gotestsum --format testname --junitfile junit-win-report.xml -- -v ./ ...
  | + ~~~~~~~~~
  | + CategoryInfo          : ObjectNotFound: (gotestsum:String) [], CommandNotFoundException
  | + FullyQualifiedErrorId : CommandNotFoundException
```
I think the installed dependencies are not detected in whatever powershell's $PATH is

leehinman (approved):
LGTM

PR #152, opened 2024-05-13
Use older version of docker that doesn't break beats

fearful-symmetry:
## What does this PR do?

So, when I added the docker container tests, I didn't realize that the newest version of the docker API is completely different, and once I imported it into beats, the newer version of the docker library just broke everything. I figure the easiest way to deal with this is just to use the same version of docker here and in beats.


cmacknz:
If there is a CVE reported against the current docker patch version we are going to be doing this work anyway. It may be better to just update the types downstream now, otherwise someone with no context on the problem might end up having to deal with as a result of a vulnerability scan result.

cmacknz:
How much work is fixing up Beats? Is it just a bunch of find/replace work? If so I'd say just do it.

If this involves rewriting actual logic then create a separate issue and we'll merge this and do the changes separately.

fearful-symmetry:
It looks like mostly library/type name changes, but there's a ton of them. Guess I'll take a crack at it and find out.

fearful-symmetry:
@cmacknz so, it looks like all our _other_ `elastic-agent-*` libraries are also on `24.x.x`, and looking at git, that branch is still getting updates as of two months ago, so I assume we'd get any fixes for CVEs on that branch as well. I think we might be better off sticking with 24.x.x for now, unless we have a reason to update the entire ecosystem of helper libs.

rdner (approved):

PR #165, opened 2024-07-17
[system/process] add support for mutlierr

VihasMakwana:
- Enhancement
## Checklist

## Author's Checklist

- [ ]

## Related issues

-


metric/system/process/process.go
```diff
@@ -27,6 +27,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/joeshaw/multierror"
```
cmacknz, line 30: https://pkg.go.dev/errors supports wrapping multiple errors now, we don't need a new dependency for this. See https://pkg.go.dev/errors#Join

VihasMakwana:
Closing this in favour of https://github.com/elastic/elastic-agent-system-metrics/pull/166

metric/system/process/process.go
VihasMakwana: Got it! Thanks for pointing this out.

PR #167, opened 2024-07-23
fix: move non-fatal error to helpers

VihasMakwana:
- Bug
## What does this PR do?

A quick fix PR to move `NonFatalErr` to helpers.go instead of process.go.
This makes sure that `NonFatalErr` compiles for all OS'.
## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/metric/system/process/helpers.go
+++ b/metric/system/process/helpers.go
@@ -99,6 +99,33 @@ func GetProcCPUPercentage(s0, s1 ProcState) ProcState {
 
 }
 
+// NonFatalErr is returned when there was an error
+// collecting metrics, however the metrics already
+// gathered and returned are still valid.
+// This error can be safely ignored, this will result
+// in having partial metrics for a process rather than
+// no metrics at all.
+//
+// It was introduced to allow for partial metrics collection
+// on privileged process on Windows.
+type NonFatalErr struct {
+	Err error
+}
+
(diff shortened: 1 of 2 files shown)


metric/system/process/helpers.go
```diff
@@ -99,6 +99,33 @@ func GetProcCPUPercentage(s0, s1 ProcState) ProcState {
...
 
+// NonFatalErr is returned when there was an error
+// collecting metrics, however the metrics already
+// gathered and returned are still valid.
+// This error can be safely ignored, this will result
+// in having partial metrics for a process rather than
+// no metrics at all.
+//
+// It was introduced to allow for partial metrics collection
+// on privileged process on Windows.
```
AndersonQ, line 110: [Suggestion]
It feels a bit out of place a type state when it's returned, therefore I'd suggest:

```suggestion
// NonFatalErr indicates an error occurred during metrics
// collection, however the metrics already
// gathered and returned are still valid.
// This error can be safely ignored, this will result
// in having partial metrics for a process rather than
// no metrics at all.
//
// It was introduced to allow for partial metrics collection
// on privileged process on Windows.
```

AndersonQ (approved):

metric/system/process/helpers.go
VihasMakwana: Thank!

Issue #164, opened 2024-07-17
[metric/system/process] - return errors encountered while monitoring set of processes


VihasMakwana:
- While monitoring set of processes, we don't return them to the caller but rather log them at debug leve.
- Example of such logs are:
```json
{"log.level":"debug","@timestamp":"2024-07-16T21:04:30.148Z","message":"Non-fatal error fetching PID metrics for 30358, metrics are valid, but partial: Not enough privileges to fetch information: /io unavailable; if running inside a container, use SYS_PTRACE: error fetching IO metrics: open /proc/30358/io: permission denied","component":{"binary":"metricbeat","dataset":"elastic_agent.metricbeat","id":"system/metrics-default","type":"system/metrics"},"log":{"source":"system/metrics-default"},"log.logger":"processes","log.origin":{"file.line":268,"file.name":"process/process.go","function":"github.com/elastic/elastic-agent-system-metrics/metric/system/process.(*Stats).pidFill"},"service.name":"metricbeat","ecs.version":"1.6.0","ecs.version":"1.6.0"}
{"log.level":"debug","@timestamp":"2024-07-16T21:04:40.143Z","message":"Non-fatal error fetching PID metrics for 30357, metrics are valid, but partial: Not enough privileges to fetch information: /io unavailable; if running inside a container, use SYS_PTRACE: error fetching IO metrics: open /proc/30357/io: permission denied","component":{"binary":"metricbeat","dataset":"elastic_agent.metricbeat","id":"system/metrics-default","type":"system/metrics"},"log":{"source":"system/metrics-default"},"log.logger":"processes","log.origin":{"file.line":268,"file.name":"process/process.go","function":"github.com/elastic/elastic-agent-system-metrics/metric/system/process.(*Stats).pidFill"},"service.name":"metricbeat","ecs.version":"1.6.0","ecs.version":"1.6.0"}
[...]

PR #172, opened 2024-08-01
[system/process] add support for mutlierr 

VihasMakwana:
## What does this PR do?

- Previously, we weren't passing errors to the caller while monitoring set of processes.
    - With the recent introduction of the status reporter for metricsets, it is impossible to change the status to degraded if such errors are not passed to the caller.
- Fix this by passing errors to the caller. We also populate the process related information to our best-effort.

## Checklist

## Manual testing and general information

- See https://github.com/elastic/beats/pull/40400 for testing it on `metricbeat`
**NOTE**: 
   - **Only applicable if you're using `system/process` module**
   - Non-fatal errors are only received when you have insufficient privileges. 

Steps:
   - While receiving any error, test for nature of error
   - call `errors.Is(err, NonFatalErr{}))` on received error
      - If true, error is non-fatal and you can proceed further (metrics will be partially available, most probably insufficient privileges).
      - Else, log the error and stop execution (metrics will be empty)

Genreal info related to the changes in this PR:
   - While getting process related information, you might also receive a non-nil error.
   - Such errors come in two flavours:
       - Fatal errors: 
          - This indicates that the error was fatal (for eg. `no process found`,)
          - Caller should stop further execution if they receive fatal errors 
       - Non-fatal errors: 
          - This indicates that the error was fatal (for eg. `not enough privileges`)
          - It means that metrics are partially filled.
          - Further execution can be continued if non-fatal errors are encountered

- Closes https://github.com/elastic/elastic-agent-system-metrics/issues/164

### DIFF
--- a/metric/system/process/helper_test.go
+++ b/metric/system/process/helper_test.go
@@ -1,0 +1,76 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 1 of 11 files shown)


rdner (commented):
Due to the nature of the change, we need to perform manual tests in order to ensure the change works. Please add instructions for manual testing to ensure updating to the new version of this library would not break Beats.

VihasMakwana:
@rdner I've added the beats PR link in description

rdner (approved):

PR #176, opened 2024-08-21
fix non-fatal error tests

fearful-symmetry:
## What does this PR do?

So, a previous PR changed the behavior of the process code such that we're now returning non-fatal errors. 
However, a bunch of the tests were not updated, and are now failing. 

I'm a little worried by the fact that none of the tests failed initially, as they fail when I run them locally. Part of the problem is that the CI environment often doesn't have any processes running as other users, meaning we won't hit a lot of code that deals with permissions errors. Still, it feels like _something_ should have failed here. Still investigating. 

## Why is it important?

This is causing tests here to fail: https://github.com/elastic/beats/pull/40507

## Checklist

### DIFF
--- a/metric/system/process/process_linux_test.go
+++ b/metric/system/process/process_linux_test.go
@@ -60,7 +60,9 @@ func TestFetchOtherProcessCgroup(t *testing.T) {
 	assert.NoError(t, err, "Init")
 
 	evts, _, err := testConfig.Get()
-	assert.ErrorIs(t, err, NonFatalErr{})
+	if err != nil {
+		assert.ErrorIs(t, err, NonFatalErr{})
+	}
 	t.Logf("Got %d events", len(evts))
 }
 
@@ -192,7 +194,9 @@ func TestCgroupsBadCgroupsConfig(t *testing.T) {
 
 	// make sure we still have proc data despite cgroups errors
 	procs, _, err := testStats.Get()
-	assert.ErrorIs(t, err, NonFatalErr{})
+	if err != nil {
+		assert.ErrorIs(t, err, NonFatalErr{})
+	}
 
 	t.Logf("got %d procs", len(procs))
 	require.NotEmpty(t, procs)
--- a/metric/system/process/process_linux_test.go
+++ b/metric/system/process/process_linux_test.go
@@ -20,6 +20,7 @@
 package process
 
 import (
+	"errors"
 	"os"
 	"os/user"
 	"strconv"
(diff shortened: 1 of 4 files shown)


metric/system/process/process_linux_test.go
```diff
@@ -60,7 +61,11 @@ func TestFetchOtherProcessCgroup(t *testing.T) {
 	assert.NoError(t, err, "Init")
 
 	evts, _, err := testConfig.Get()
-	require.NoError(t, err)
+	if err != nil {
+		assert.True(t, errors.Is(err, NonFatalErr{}))
+	} else {
+		assert.NoError(t, err, "GetOne with one event")
+	}
```
AndersonQ, line 68: [Blocker]
This is always true, `if err == nil` calling `assert.NoError` is redundant.
same for the other tests.

metric/system/process/process_linux_test.go
```diff
@@ -60,7 +61,11 @@ func TestFetchOtherProcessCgroup(t *testing.T) {
 	assert.NoError(t, err, "Init")
 
 	evts, _, err := testConfig.Get()
-	require.NoError(t, err)
+	if err != nil {
+		assert.True(t, errors.Is(err, NonFatalErr{}))
```
AndersonQ, line 65: [Blocker]
It'd be better to have a message on the `assert.True`. Getting a "false is not true" as the error message ins't much of a help.

Also, in its current state we'd lose the error, making it hard to understand the issue.
same for the other tests.
```suggestion
		assert.Truef(t, errors.Is(err, NonFatalErr{}, "err isn't NonFatalErr. err: %v", err))
```

VihasMakwana:
> I'm a little worried by the fact that none of the tests failed initially, as they fail when I run them locally. Part of the problem is that the CI environment often doesn't have any processes running as other users, meaning we won't hit a lot of code that deals with permissions errors. Still, it feels like something should have failed here. Still investigating.

@fearful-symmetry I think I know. 

When running these tests locally, I used to run them as the root user, and they didn't fail under those conditions. I assumed the CI environment also runs tests as root, which is why no failures were detected there.

I addressed the unprivileged case in the beats repository [here](https://github.com/elastic/beats/blob/e09627fb56729700113dd7a626ef4472f0e1bf2c/metricbeat/module/system/process/process_test.go#L40-L42).

Thanks for taking care of it on this repo!

metric/system/process/process_linux_test.go
fearful-symmetry: Argh, meant to check for `NotFatalError`

fearful-symmetry:
@VihasMakwana yeah, a lot of the process tests are very sensitive to different permissions contexts, which is why this repo has the container tests that are meant to test a dozen different docker permissions settings.

mauri870 (approved):

PR #181, opened 2024-10-18
[windows][system/process] - Get Idle process' stats

VihasMakwana:
- Enhancement

## What does this PR do?

This PR adds two functions that captures following information for idle process (pid 0). This is the only way I came across as we cannot open process handle for pid 0.
- memory stats
- CPU usage
- number of threads

Future PR:
- In a follow up PR, I'll fix metric collection for PID 4.
- The a final PR to integrate changes and update the final process map to include PID 0 and 4.
- Then, we can merge it to main after testing it thoroughly.

## Why is it important?

- See https://github.com/elastic/beats/issues/40484 for more info

## NOTE:

I've created a dedicated branch called `windows-protected-processes` to hold all the changes. Once we made the necessary updates for protected process, I'll test them manually and then submit a final PR to merge it into the `main` branch. I think this is the best approach to avoid any issues and CI failures due to releases. Let me know your thoughts!

### DIFF
--- a/metric/system/process/process_windows.go
+++ b/metric/system/process/process_windows.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/elastic/elastic-agent-libs/opt"
 	"github.com/elastic/elastic-agent-system-metrics/metric/system/resolve"
+	gowindows "github.com/elastic/go-windows"
 	"github.com/elastic/gosigar/sys/windows"
 )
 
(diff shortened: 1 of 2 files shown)


metric/system/process/process_windows.go
```diff
@@ -351,3 +385,49 @@ func getProcCredName(pid int) (string, error) {
...
+	num_cpus := float64(runtime.NumCPU())
+	idleTime := float64(idle) / num_cpus
+	kernelTime := float64(kernel) / num_cpus
+	userTime := float64(user) / num_cpus
+
+	if err != nil {
+		return 0, 0
+	}
+	// Calculate total CPU time, averaged by cpu
+	totalTime := idleTime + kernelTime + userTime
```
leehinman, line 402: I'm not sure if this is right.  The following doc makes me believe that kernelTime includes idleTime.  Did you find this to be different in practice?

 https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getsystemtimes#parameters

VihasMakwana: You're correct.kernelTime includes idleTime If I directly use the win32 api.

But as you can see here, I use `https://github.com/elastic/go-windows`
 https://github.com/elastic/elastic-agent-system-metrics/blob/5e3c5cf781aadb14be48e4ef2216e292b77fcdd2/metric/system/process/process_windows.go#L390

 and that takes the above thing into the account and calculates kernelTime as `kernelTime - idleTime` before returning the values. See [here](https://github.com/elastic/go-windows/blob/ee94ac30503b3ecb77e35710f8648d500919f3b9/kernel32.go#L192-L204)

VihasMakwana: The results of this function are accurate with the task manager's stats.

leehinman: Great, Thanks.

leehinman (approved):

PR #184, opened 2024-10-22
[chore][windows][system/process]: ignore error in some cases

VihasMakwana:
- Enhancement

We can ignore the error in two cases:
- While reading the process executable name. 
    - For pid 4, this call fails as we can't access the executable name via the system call. Same for other kernel-level processes.
- While finding the owner for a particular process.
    - We try to open the process token via `syscall.OpenProcessToken`and we can't access the token for protected processes , even as an administrator. 
    
it's okay to ignore these errors and move forward as we can access few other metrics (memory, cpu).
    
More context [here](https://github.com/elastic/beats/issues/40484#issuecomment-2400100817)

Relates https://github.com/elastic/beats/issues/40484

### DIFF
--- a/metric/system/process/process_windows.go
+++ b/metric/system/process/process_windows.go
@@ -265,6 +265,8 @@ func getProcName(pid int) (string, error) {
 	}()
 
 	filename, err := windows.GetProcessImageFileName(handle)
+
+	//nolint:nilerr // safe to ignore this error
 	if err != nil {
 		// if we're able to open the handle but GetProcessImageFileName fails then it most probably means
 		// that the process doesn't have any executable associated with it.
--- a/metric/system/process/process_windows.go
+++ b/metric/system/process/process_windows.go
@@ -136,8 +136,9 @@ func FillPidMetrics(_ resolve.Resolver, pid int, state ProcState, _ func(string)
 	user, _ := getProcCredName(pid)
 	state.Username = user // we cannot access process token for system-owned protected processes
 
-	ppid, _ := getParentPid(pid)
-	state.Ppid = opt.IntWith(ppid)
+	if ppid, err := getParentPid(pid); err == nil {
+		state.Ppid = opt.IntWith(ppid)
+	}
 
 	wss, size, err := procMem(pid)
 	if err != nil {
--- a/metric/system/process/process_windows.go
+++ b/metric/system/process/process_windows.go
@@ -133,11 +133,8 @@ func FetchNumThreads(pid int) (int, error) {
 
 // FillPidMetrics is the windows implementation
 func FillPidMetrics(_ resolve.Resolver, pid int, state ProcState, _ func(string) bool) (ProcState, error) {
-	user, err := getProcCredName(pid)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


metric/system/process/process_windows.go
```diff
@@ -133,11 +133,8 @@ func FetchNumThreads(pid int) (int, error) {
...
-	user, err := getProcCredName(pid)
-	if err != nil {
-		return state, fmt.Errorf("error fetching username: %w", err)
-	}
-	state.Username = user
+	user, _ := getProcCredName(pid)
+	state.Username = user // we cannot access process token for system-owned protected processes
 
 	ppid, _ := getParentPid(pid)
 	state.Ppid = opt.IntWith(ppid)
```
leehinman, line 140: So if we don't error and return, then we are going to call getParentPid, which I think would return 0 for the ppid (assuming this will fail like getProcCredName).  That seems like a problem, we don't really know if the ppid is 0 or not.

I guess the question I have is.  Is the state that is returned when getProcCredName fails accurate?

metric/system/process/process_windows.go
```diff
@@ -268,8 +265,12 @@ func getProcName(pid int) (string, error) {
 	}()
 
 	filename, err := windows.GetProcessImageFileName(handle)
+
+	//nolint:nilerr // safe to ignore this error
 	if err != nil {
-		return "", fmt.Errorf("GetProcessImageFileName failed for pid=%v: %w", pid, err)
+		// if we're able to open the handle but GetProcessImageFileName fails then it most probably means
+		// that the process doesn't have any executable associated with it.
```
leehinman, line 272: Could we detect this based on the error?  I'd rather not swallow the error if it is possible it might contain useful information, especially when it isn't the common case.

metric/system/process/process_windows.go
VihasMakwana: > Is the state that is returned when getProcCredName fails accurate

Not really. The current flow of this function is:
- Get Process username, return error if failed
- Get Parent PID, ignore error if encountered
- Get memory metrics, return error if failed 
- Get CPU metrics, return error if failed

For protected processes, we fail at 1st step (we try to get the token for a protected process, which is not allowed). But, we can access memory and CPU metrics.

So, the state returned when getProcCredName fails is not accurate as we miss out on certain metrics later.

metric/system/process/process_windows.go
VihasMakwana: Yes. We can ignore the error if it's Access Denied or Operation not Permitted. I'll update the PR.

metric/system/process/process_windows.go
leehinman: I'm worried about when/if getParentPid fails.  If it fails, we get 0 for the ppid.  My guess is that getParentPid would always succeed when getProcCredName succeeded.  So this would be a new case where we are reporting the ppid incorrectly.

VihasMakwana: @leehinman I've found a solution to this. Let me know how you feel about this.

mauri870 (approved):

leehinman (approved):
LGTM

PR #185, opened 2024-10-23
[system/process] - Add idle process' stats

VihasMakwana:
This PR does the following:
- It integrates `getIdleMemory` and `getProcessMemory` introduced in https://github.com/elastic/elastic-agent-system-metrics/pull/181. This is used to capture stats for PID 0.
- `FillMetricsRequiringMoreAccess` adds some extra information about a process. It's failure is expected for protected processes and if a non-root user accesses elevated processes. We can just log the error rather than returning an error. This will not pollute the status message on fleet.
     - @cmacknz Please let me know how you feel about this.

This is final PR in this series and once this PR gets merged, we can merge `windows-protected-processes` into `main`.

### DIFF
--- a/metric/system/process/process_windows.go
+++ b/metric/system/process/process_windows.go
@@ -463,12 +463,16 @@ func fillIdleProcess(state ProcState) (ProcState, error) {
 	if err != nil {
 		return state, err
 	}
-	time.Sleep(500 * time.Millisecond)
+	// sleep is cruical to calculate the idle percentage
+	time.Sleep(50 * time.Millisecond)
 	total1, idle1, err := getIdleProcessTime()
 	if err != nil {
 		return state, err
 	}
-	state.CPU.Total.Pct = opt.FloatWith((idle1 - idle0) / (total1 - total0))
+	// Check if denominator is non-zero, to avoid "invalid operation: division by zero"
+	if total1-total0 != 0 {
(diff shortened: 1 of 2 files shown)


cmacknz:
> We can just log the error rather than returning an error. This will not pollute the status message on fleet.

Making this a debug log feels reasonable to me.

metric/system/process/process_windows.go
```diff
@@ -436,3 +453,23 @@ func getIdleProcessMemory(state ProcState) (ProcState, error) {
...
+func fillIdleProcess(state ProcState) (ProcState, error) {
+	state, err := getIdleProcessMemory(state)
+	if err != nil {
+		return state, err
+	}
+	total0, idle0, err := getIdleProcessTime()
+	if err != nil {
+		return state, err
+	}
+	time.Sleep(500 * time.Millisecond)
```
leehinman, line 466: Do we have to Sleep?  I might be missing something but I think this means we will wait 1/2 sec every time we try to get process stats on windows, because PID 0 will always be in the list.

VihasMakwana: Well, I'm doing this to calculate the percentage in which CPU was idle. 
I think we can reduce the timeout. 1/2 second is bit too much.

VihasMakwana: @leehinman i've used `50s` as a timeout value.

leehinman (approved):

PR #188, opened 2024-10-28
update to go 1.22

VihasMakwana:
`beats` and `elastic-agent` are already on go 1.22. Update to 1.2.

This also fixes [CI failure](https://buildkite.com/elastic/elastic-agent-system-metrics/builds/438#0192d318-7456-49d1-8ccb-bf9f1fd6febb)


.go-version
```diff
@@ -1 +1 @@
-1.21.13
+1.22.0
```
belimawr, line 1: The `.go-version` from Beats and Elastic-Agent is `1.22.8`, so I suggest using the same here.
```suggestion
1.22.8
```

rdner:
@rowlandgeoff should we setup the automation here too, like we have in other repositories?

rdner:
@VihasMakwana please update versions in this file too https://github.com/elastic/elastic-agent-system-metrics/blob/main/.golangci.yml

rowlandgeoff:
> @rowlandgeoff should we setup the automation here too, like we have in other repositories?

We chatted about this a bit in the Eng Prod office hours meeting today. The summary was that we definitely need more consistency in how we do version bumps across the repos. Various repos use dependabot, GH actions, or renovate. 

This repo does have a GH action that should be bumping the version.
- https://github.com/elastic/elastic-agent-system-metrics/actions/workflows/bump-golang.yml
- https://github.com/elastic/elastic-agent-system-metrics/blob/main/.github/workflows/bump-golang.yml
- https://github.com/elastic/elastic-agent-system-metrics/blob/main/.github/workflows/updatecli.d/bump-golang.yml

Looking at the [comparable history in elastic-agent](https://github.com/elastic/elastic-agent/commits/main/.go-version), it looks like the GH action will bump the patch versions, but it will not bump the minor version. That part seems to be done manually.

belimawr (approved):

belimawr (commented):
LGTM, but @rdner comment still needs to be addressed before merging.

rdner (approved):

PR #186, opened 2024-10-28
[system/process][darwin] return error as non-fatal if invalid env k-v pair is encountered

VihasMakwana:
## What does this PR do?

- Implements one of the solutions for https://github.com/elastic/beats/issues/41461#issuecomment-2441387003

## Why is it important?

- See https://github.com/elastic/beats/issues/41461#issuecomment-2441387003

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `CHANGELOG.md`

## Related issues
- Closes https://github.com/elastic/beats/issues/41461

## Steps to reproduce
- See https://github.com/elastic/beats/issues/41461#issuecomment-2441387003

PR #189, opened 2024-11-05
[bugfix][system/process] continue if we encounter an error, or it'll lead to panic

VihasMakwana:
continue if we encounter an error, or it'll lead to panic here https://github.com/elastic/elastic-agent-system-metrics/blob/ef8d2d0a7e9e0a64a427b190b8c113141f1a99dc/metric/system/process/process_darwin.go#L240-L242
 
 We return an error if we env pairs are invalid. See https://github.com/elastic/elastic-agent-system-metrics/pull/186.
 `Continue` if we face error.


mauri870 (approved):
LGTM

cmacknz:
Can we add a unit test (in a separate PR) to cover this case so we don't accidentally regress on it later?

VihasMakwana:
@cmacknz sure. I'll add it

PR #192, opened 2024-11-18
[cpu/windows] Switch to performance counters

VihasMakwana:
Add conditional switching to performance counters for `cpu`/`core` metricset collections
This PR add an option `WithPerformanceCounter`, which uses performance counter for CPU data collection, when enabled. 
Default behaviour will be the existing implementation.

### How to use performance counters?

It's simple. Something like following:
```go
cpu, _ := metrics.New(WithPerformanceCounter())
data, err := cpu.Fetch()
```

### Why performance counters?
From https://github.com/elastic/beats/issues/40926,
> On Windows, Metricbeat measures CPU use via the Windows API call GetSystemTimes. Each metrics interval, it fetches the CPU numbers, and compares them to the previous measurement to determine CPU load during that interval. On most systems this includes CPU time ["including all threads in all processes, on all processors"](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getsystemtimes). However, on systems with more than 64 cores, it returns only the data for the [current processor group](https://learn.microsoft.com/en-us/windows/win32/procthread/processor-groups) of up to 64 cores.

This is a major limitation for such systems and it leads to inconsistent data.

### Testing

Test results are reported on the linked issue.

Fixes https://github.com/elastic/beats/issues/40926

### **Note:**
`WithPerformanceCounter` option is only effective for windows and is ineffective if used by other OS'.


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.22.8
 
 require (
 	github.com/docker/docker v26.1.5+incompatible
-	github.com/elastic/elastic-agent-libs v0.9.13
+	github.com/elastic/elastic-agent-libs v0.17.3-0.20241112062438-5ba501c7ca8c
```
VihasMakwana, line 7: This will be updated once https://github.com/elastic/elastic-agent-libs/pull/252 is merged and released

pierrehilbert:
`Error: some files are not up-to-date. Usually running 'mage update' or 'mage addLicenseHeaders' fixes the issues. Fix the issues, review and commit the changes. Modified: [go.sum]`

metric/cpu/metrics_windows.go
```diff
@@ -24,45 +24,200 @@ package cpu
...
+var (
+	processorInformationCounter = "\\Processor Information(%s)\\%s"
+	totalKernelTimeCounter      = fmt.Sprintf(processorInformationCounter, "_Total", "% Privileged Time")
+	totalIdleTimeCounter        = fmt.Sprintf(processorInformationCounter, "_Total", "% Idle Time")
+	totalUserTimeCounter        = fmt.Sprintf(processorInformationCounter, "_Total", "% User Time")
+)
+
+var (
+	// a call to getAllCouterPaths is idempodent i.e. it returns same set of counters every time you call it.
+	// we can save some cruicial cycles by converting it to a sync.Once
```
mauri870, line 46:
```suggestion
	// we can save some crucial cycles by converting it to a sync.Once
```

metric/cpu/metrics_windows.go
```diff
@@ -24,45 +24,200 @@ package cpu
...
+	getAllCouterPathsOnce = sync.OnceValues(getAllCouterPaths)
```
mauri870, line 47: It's best to search for couter to ensure every occurrence is corrected.

metric/cpu/metrics_windows.go
```diff
@@ -24,45 +24,200 @@ package cpu
...
+
+		if _, ok := cpuMap[instance]; !ok {
+			cpuMap[counter.instance] = &CPU{}
+		}
+		val, err := q.GetRawCounterValue(name)
+		if err != nil {
+			return nil, fmt.Errorf("call to GetRawCounterValue failed for %s: %w", counter, err)
+		}
+		// the counter value returned by GetRawCounterValue is in 100-ns intervals
+		// convert it to nanoseconds
```
mauri870, line 126: Should this be `convert it to milliseconds?`

mauri870 (approved):
Aside from some minor issues, it looks good to me. I'm not very experienced with Windows, so it's better to get a review from a more experienced developer on the topic.

cmacknz:
The most important thing here is that we did not break or unnecessarily change the meaning of the metrics from what they were before.

How did we test this? How did we verify that it solves https://github.com/elastic/beats/issues/40926? Are the existing automated tests good enough to regression test this change?

VihasMakwana:
> The most important thing here is that we did not break or unnecessarily change the meaning of the metrics from what they were before.
> 
> How did we test this? How did we verify that it solves [elastic/beats#40926](https://github.com/elastic/beats/issues/40926)? Are the existing automated tests good enough to regression test this change?

I agree. Currently, I have manually tested and compared the values against the old implementation. They check out.
For my next task, I'll compare values on kibana and share the results with the team.

> How did we verify that it solves [elastic/beats#40926](https://github.com/elastic/beats/issues/40926)?

We have a GCP VM with 90 cores. 
If we rely on existing implementation, then we only get data for 64 cores (current processor group). But, using the `Processor Information` we can get data for all the cores installed on the system. As mentioned, I'll run `elastic-agent` with this fix on this VM and compare/share results.

VihasMakwana:
Will mark it for review once I'm done with the benchmarking.

cmacknz:
Let's make this a change we can opt in and out of via configuration in the places that that will consume this change. https://github.com/elastic/beats/issues/40926#issuecomment-2501257448

I think this is actually more places than just the system/metrics input in beats, since I think elastic-agent and beats both use it to expose metrics via their /stats pages.

metric/cpu/metrics.go
```diff
@@ -84,6 +84,18 @@ The below code implements a "metrics tracker" that gives us the ability to
 calculate CPU percentages, as we average usage across a time period.
 */
 
+type option struct {
+	usePerformanceCounter bool
+}
+
+type OptionFunc func(*option)
+
+func WithPerformanceCounter() OptionFunc {
```
mauri870, line 93: I assume performance counters is just a windows thing? If so, we could add a comment at the top of this function that states that it only applies to windows.

metric/cpu/metric_windows_test.go
```diff
@@ -0,0 +1,51 @@
...
+package cpu
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCounterLength(t *testing.T) {
+	query, err := buildQuery()
```
mauri870, line 29: Isn't it better to rely on `var query, qError = buildQuery()` instead of calling buildQuery explicitly? This way we make sure that the globals used by the real library code works as intended. This guards against any issue that can happen during initialization. Does that make sense here?

mauri870 (approved):
Reviewed again, LGTM. I'm learning quite a bit about Windows internals through your PR chain.

metric/cpu/metric_windows_test.go
mauri870: Additionally, I'm not sure if this has any impact but wouldn't be better to call buildQuery inside init()? I think globals are initialized before init, but I might be wrong.

VihasMakwana: > Additionally, I'm not sure if this has any impact but wouldn't be better to call buildQuery inside init()

That would look better. I'll update the PR. Thank!

> Isn't it better to rely on var query, qError = buildQuery() instead of calling buildQuery explicitly

I'll move the `var query, qError = buildQuery()` under `init()`.

metric/cpu/metrics.go
cmacknz: Putting Windows in the option name would be even more obvious.

Also, consider that we are probably going to want to make this the default once we are satisfied with it. Probably just deleting the option would work once we do that but something to consider for the near future.

metric/cpu/metrics_windows.go
```diff
@@ -26,13 +26,102 @@ import (
...
 	"time"
 
+	"github.com/elastic/elastic-agent-libs/helpers/windows/pdh"
 	"github.com/elastic/elastic-agent-libs/opt"
 	"github.com/elastic/elastic-agent-system-metrics/metric/system/resolve"
 	"github.com/elastic/gosigar/sys/windows"
 )
 
+var (
+	processorInformationCounter = "\\Processor Information(%s)\\%s"
```
cmacknz, line 36: Doing some searching, I think these names might be localized to the language of the Windows installs. See https://stackoverflow.com/questions/5581859/retrieve-performance-counter-value-in-a-language-independent-way for example. It looks like the more common ones can be identified by an ID instead of a locale dependent string.

Let's make sure we test this on a non-english version of Windows. We have had problems in the past on the French and German language installs of Windows.

See https://github.com/elastic/elastic-agent/issues/3960#issuecomment-1877700201 for example to get started.

VihasMakwana: Yup. I recently had same discussion with @leehinman on https://github.com/elastic/elastic-agent-libs/pull/254#discussion_r1856899234.

I updated the functions to add counters in language neutral manner. See [here](https://github.com/VihasMakwana/elastic-agent-libs/blob/0d290c679f51bcf5f7d74a42ddf8e36a4ab4f878/helpers/windows/pdh/pdh_query_windows.go#L98-L103).

> Let's make sure we test this on a non-english version of Windows. We have had problems in the past on the French and German language installs of Windows.

Will do it. Thanks!

metric/cpu/metrics_windows.go
```diff
@@ -26,13 +26,102 @@ import (
...
+	totalKernelTimeCounter      = fmt.Sprintf(processorInformationCounter, "*", "% Privileged Time")
+	totalIdleTimeCounter        = fmt.Sprintf(processorInformationCounter, "*", "% Idle Time")
+	totalUserTimeCounter        = fmt.Sprintf(processorInformationCounter, "*", "% User Time")
+)
+
+var query pdh.Query
+var qError error
+
+func init() {
+	query, qError = buildQuery()
```
cmacknz, line 46: We are throwing away the error here, is there a better way to initialize this? Perhaps a  or sync.Once or something?

If the counter names are in fact localized, this is where the error would first appear, and we'd never see it with the code the way it is now.

VihasMakwana: sync.Once adds CPU overhead. I'll explore more options.

> If the counter names are in fact localized, this is where the error would first appear, and we'd never see it with the code the way it is now.

I do check for qError at https://github.com/elastic/elastic-agent-system-metrics/blob/35a0aa417f52b238bb1b9f4926465e3b61758e41/metric/cpu/metrics_windows.go#L50-L61.

VihasMakwana: We will pass the error to the caller and eventually to metricbeat.

metric/cpu/metrics_windows.go
cmacknz: Nice! Thank you @leehinman for also being haunted by this problem.

metric/cpu/metrics_windows.go
cmacknz: If we have to add an initializer function for system metrics it's not the end of the world. 

Looks like there are a couple of other instances of this problem in here already, like https://github.com/elastic/elastic-agent-system-metrics/blob/5a3873d8d6f807e2c111b437c10291a13d9c2b83/report/metrics_common.go#L38-L47

Although this is the first one where we have a real reason for it to fail (thanks Windows localization).

leehinman: Please don't do this in `init()`, if there are any delays or blocking look ups that can delay startup of elastic-agent.  That is a problem for things like running as a Windows service where we have a set amount of time to respond to the Service Manager.

leehinman (changes requested):

metric/cpu/metrics_windows.go
VihasMakwana: @leehinman I've opened another, smaller PR https://github.com/elastic/elastic-agent-system-metrics/pull/194 just for handling this. Can you take a look?

leehinman: ✅ That is a big improvement.  Thank You.

metric/cpu/metrics_windows.go
```diff
@@ -27,12 +27,68 @@ import (
...
+	totalIdleTimeCounter        = fmt.Sprintf(processorInformationCounter, "*", "% Idle Time")
+	totalUserTimeCounter        = fmt.Sprintf(processorInformationCounter, "*", "% User Time")
+)
+
 // Get fetches Windows CPU system times
-func Get(_ resolve.Resolver) (CPUMetrics, error) {
+func Get(m *Monitor) (CPUMetrics, error) {
+	if m.query == nil {
+		return defaultGet()
+	}
```
leehinman, line 44: Nit.  Just to improve readability and testing.  I think it would be nice if `Get` was something like:

```
func Get(m *Monitor) (CPUMetrics, error) {
    if m.query == nil {
        return getUsingSystemTimes()
    }
    return getUsingPerformanceCounters(m)
}
```

That way it is explicit that the only thing Get does is pick which implementation to use, and it is easy to test the 2 implementations independently.

leehinman (approved):
LGTM, optional refactor.

metric/cpu/metrics_windows.go
VihasMakwana: Got it. Thanks!

cmacknz (approved):

PR #194, opened 2024-12-03
[cpu] - Split the initialisation into multiple files

VihasMakwana:
This PR is a split from https://github.com/elastic/elastic-agent-system-metrics/pull/192. 
It aims to simplify the query building for performance counters.

Using a global variable might create problems if multiple modules (`cpu` and `core`) are trying to initialise it at a same time.
`sync.Once` adds unnecessary CPU overhead (makes it 4x slower). 

Instead, we can initialise the performance counters in `New(...)` and return error to the caller (i.e. metricbeat).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.22.8
 
 require (
 	github.com/docker/docker v26.1.5+incompatible
-	github.com/elastic/elastic-agent-libs v0.9.13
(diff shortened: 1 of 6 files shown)


leehinman (approved):
Thank You for getting the query out of `init()`.  This is much better.

PR #198, opened 2024-12-17
[process/windows] - Ignore accessing command line arguments for selected processes (currently, lsass.exe)

VihasMakwana:
See https://github.com/elastic/beats/issues/41407 for more details.

TL;DR;

Processes such as `lsass.exe` has no meaningful cmd line and and can trigger false positives with Windows ASR rules.
We can find pid for that using `SYSTEM\\CurrentControlSet\\Control\\Lsa` path

Relates https://github.com/elastic/beats/issues/41407

### DIFF
--- a/metric/system/process/helpers_windows.go
+++ b/metric/system/process/helpers_windows.go
@@ -23,9 +23,10 @@ import (
 	"errors"
 	"syscall"
 
-	"github.com/elastic/elastic-agent-libs/logp"
 	"golang.org/x/sys/windows"
 	"golang.org/x/sys/windows/registry"
+
+	"github.com/elastic/elastic-agent-libs/logp"
 )
 
 func isNonFatal(err error) bool {
--- a/metric/system/process/helpers_windows.go
+++ b/metric/system/process/helpers_windows.go
@@ -39,7 +39,8 @@ func isNonFatal(err error) bool {
 		errors.Is(err, windows.ERROR_INVALID_PARAMETER) || errors.Is(err, NonFatalErr{})
 }
 
-func processesToIgnore() (m map[uint64]struct{}) {
+func processesToIgnore() map[uint64]struct{} {
+	m := make(map[uint64]struct{})
 	// processesToIgnore checks if we should ignore the pid, to avoid elevated permissions
 
 	// LSASS.exe is a process which has no useful cmdline arguments, we should ignore acessing such process to avoid triggering Windows ASR rules
--- a/metric/system/process/helpers_windows.go
+++ b/metric/system/process/helpers_windows.go
@@ -55,5 +55,5 @@ func processesToIgnore() (m map[uint64]struct{}) {
 		return m
 	}
 	m[lsassPid] = struct{}{}
-	return
+	return m
 }
--- a/metric/system/process/helpers_windows.go
+++ b/metric/system/process/helpers_windows.go
@@ -23,7 +23,9 @@ import (
 	"errors"
 	"syscall"
 
+	"github.com/elastic/elastic-agent-libs/logp"
 	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
 )
 
 func isNonFatal(err error) bool {
@@ -34,4 +36,24 @@ func isNonFatal(err error) bool {
 		errors.Is(err, syscall.EPERM) ||
 		errors.Is(err, syscall.EINVAL) ||
 		errors.Is(err, windows.ERROR_INVALID_PARAMETER) || errors.Is(err, NonFatalErr{})
+}
+
+func processesToIgnore() (m map[uint64]struct{}) {
+	// shouldIgnore checks if we should ignore the pid, to avoid elevated permissions
+
+	// LSASS.exe is a process which has no useful cmdline arguments, we should ignore acessing such process to avoid triggering Windows ASR rules
+	// we can query pid for LASASS.exe from registry
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, "SYSTEM\\CurrentControlSet\\Control\\Lsa", registry.READ)
+	if err != nil {
+		logp.L().Warnw("Failed to read registry path SYSTEM\\CurrentControlSet\\Control\\Lsa", "error", err)
+		return m
+	}
+	defer key.Close()
+	lsassPid, _, err := key.GetIntegerValue("LasPid")
+	if err != nil {
+		logp.L().Warnw("Failed to read pid for lsass.exe", "error", err)
+		return m
+	}
+	m[lsassPid] = struct{}{}
+	return
 }
--- a/metric/system/process/helpers_windows.go
+++ b/metric/system/process/helpers_windows.go
@@ -47,7 +47,7 @@ func processesToIgnore() map[uint64]struct{} {
 	// we can query pid for LASASS.exe from registry
 	key, err := registry.OpenKey(registry.LOCAL_MACHINE, "SYSTEM\\CurrentControlSet\\Control\\Lsa", registry.READ)
 	if err != nil {
-		logp.L().Warnw("Failed to read registry path SYSTEM\\CurrentControlSet\\Control\\Lsa", "error", err)
+		logp.L().Warnw("Failed to open registry path SYSTEM\\CurrentControlSet\\Control\\Lsa", "error", err)
 		return m
 	}
 	defer key.Close()
--- a/metric/system/process/helpers_windows.go
+++ b/metric/system/process/helpers_windows.go
@@ -39,7 +39,7 @@ func isNonFatal(err error) bool {
 }
 
 func processesToIgnore() (m map[uint64]struct{}) {
-	// shouldIgnore checks if we should ignore the pid, to avoid elevated permissions
+	// processesToIgnore checks if we should ignore the pid, to avoid elevated permissions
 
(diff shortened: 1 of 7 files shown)


metric/system/process/process_windows.go
```diff
@@ -463,3 +466,25 @@ func fillIdleProcess(state ProcState) (ProcState, error) {
...
+	// shouldIgnore checks if we should ignore the pid, to avoid elevated permissions
+
+	// LSASS.exe is a process which has no useful cmdline arguments, we should ignore acessing such process to avoid triggering Windows ASR rules
+	// we can query pid for LASASS.exe from registry
+
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, "SYSTEM\\CurrentControlSet\\Control\\Lsa", registry.READ)
+	if err != nil {
+		logp.L().Warnw("Failed to read registry path SYSTEM\\CurrentControlSet\\Control\\Lsa", "error", err)
+		return false
+	}
```
andrewkroh, line 482: To avoid leaking handles: `defer key.Close()`

metric/system/process/process_windows.go
```diff
@@ -199,6 +199,9 @@ func FillMetricsRequiringMoreAccess(pid int, state ProcState) (ProcState, error)
 }
 
 func getProcArgs(pid int) ([]string, error) {
+	if ok := shouldIgnore(pid); ok {
+		return []string{}, nil
```
andrewkroh, line 205: Do we need to create a non-nil empty slice? Can we just return `nil`?

https://go.dev/wiki/CodeReviewComments#declaring-empty-slices

VihasMakwana: Oh, missed this. Returning `nil` is better.

metric/system/process/process_windows.go
VihasMakwana: Thank!

metric/system/process/process_windows.go
```diff
@@ -463,3 +468,26 @@ func fillIdleProcess(state ProcState) (ProcState, error) {
...
 	return state, nil
 }
+
+func shouldIgnore(pid int) bool {
+	// shouldIgnore checks if we should ignore the pid, to avoid elevated permissions
+
+	// LSASS.exe is a process which has no useful cmdline arguments, we should ignore acessing such process to avoid triggering Windows ASR rules
+	// we can query pid for LASASS.exe from registry
+
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, "SYSTEM\\CurrentControlSet\\Control\\Lsa", registry.READ)
```
cmacknz, line 478: Are there any performance concerns doing this unconditionally for every PID in the system?

This could probably happen once right before iterating over every PID and have a low chance of being wrong, unless we know this PID changes all the time.

https://github.com/elastic/elastic-agent-system-metrics/blob/5a3873d8d6f807e2c111b437c10291a13d9c2b83/metric/system/process/process_windows.go#L51-L57

VihasMakwana: @cmacknz yeah, I think moving this to global would solve this. PIDs of such process shouldn't change.

metric/system/process/helpers_windows.go
```diff
@@ -35,3 +38,23 @@ func isNonFatal(err error) bool {
...
 		errors.Is(err, windows.ERROR_INVALID_PARAMETER) || errors.Is(err, NonFatalErr{})
 }
+
+func processesToIgnore() (m map[uint64]struct{}) {
+	// processesToIgnore checks if we should ignore the pid, to avoid elevated permissions
+
+	// LSASS.exe is a process which has no useful cmdline arguments, we should ignore acessing such process to avoid triggering Windows ASR rules
+	// we can query pid for LASASS.exe from registry
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, "SYSTEM\\CurrentControlSet\\Control\\Lsa", registry.READ)
+	if err != nil {
```
swiatekm, line 48: Under what circumstances do we get an error here?

VihasMakwana: If you're running the elastic-agent in unprivileged mode and the `elastic-agent-user` doesn't have enough permissions to read the registry.

VihasMakwana: or if the key is not present (which is highly unlikely in this scenario)

swiatekm: In that case we just emit a warning and fall back to previous behaviour though, so no risk of breaking a currently working installation via this change.

metric/system/process/process.go
```diff
@@ -291,15 +291,17 @@ func (procStats *Stats) pidFill(pid int, filter bool) (ProcState, bool, error) {
...
 
-	// Generate `status.Cmdline` here for compatibility because on Windows
-	// `status.Args` is set by `FillMetricsRequiringMoreAccess`.
-	if len(status.Args) > 0 && status.Cmdline == "" {
-		status.Cmdline = strings.Join(status.Args, " ")
+		// Generate `status.Cmdline` here for compatibility because on Windows
+		// `status.Args` is set by `FillMetricsRequiringMoreAccess`.
+		if len(status.Args) > 0 && status.Cmdline == "" {
+			status.Cmdline = strings.Join(status.Args, " ")
+		}
```
swiatekm, line 304: Could we add a unit test for this code? An integration test on Linux with `excludedPids` manually set would work as well.

swiatekm (commented):
Looks good, I'd love to see at least one unit test for the new behaviour.

metric/system/process/helpers_windows.go
VihasMakwana: @swiatekm that's what we do. See here https://github.com/elastic/elastic-agent-system-metrics/blob/33fcaec96311c1b468e7fb02284b71740b5257e9/metric/system/process/helpers_windows.go#L48-L51

Just log the warning and return.

metric/system/process/helpers_windows.go
```diff
@@ -35,3 +38,23 @@ func isNonFatal(err error) bool {
...
+		logp.L().Warnw("Failed to read registry path SYSTEM\\CurrentControlSet\\Control\\Lsa", "error", err)
+		return m
+	}
+	defer key.Close()
+	lsassPid, _, err := key.GetIntegerValue("LasPid")
+	if err != nil {
+		logp.L().Warnw("Failed to read pid for lsass.exe", "error", err)
+		return m
+	}
+	m[lsassPid] = struct{}{}
```
mauri870, line 58: I'm suspicious this will trigger a panic with `assignment to entry in nil map`. Can you confirm that?

metric/system/process/process.go
VihasMakwana: On it.

metric/system/process/helpers_windows.go
VihasMakwana: Fixed this

metric/system/process/helpers_windows.go
```diff
@@ -35,3 +38,24 @@ func isNonFatal(err error) bool {
...
+
+	// LSASS.exe is a process which has no useful cmdline arguments, we should ignore acessing such process to avoid triggering Windows ASR rules
+	// we can query pid for LASASS.exe from registry
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, "SYSTEM\\CurrentControlSet\\Control\\Lsa", registry.READ)
+	if err != nil {
+		logp.L().Warnw("Failed to read registry path SYSTEM\\CurrentControlSet\\Control\\Lsa", "error", err)
+		return m
+	}
+	defer key.Close()
+	lsassPid, _, err := key.GetIntegerValue("LasPid")
```
gabriellandau, line 54: This would never have worked.  Can we add a unit test for this?

```suggestion
	lsassPid, _, err := key.GetIntegerValue("LsaPid")
```

gabriellandau (changes requested):

metric/system/process/helpers_windows.go
VihasMakwana: Ah, that’s why it wasn’t working. I double-checked everything but forgot about the key. I’m working on the unit test case and will update it with the correct key. Thanks for pointing that out!

metric/system/process/process_windows.go
andrewkroh: > yeah, I think moving this to global would solve this. PIDs of such process shouldn't change.

@gabriellandau Can the lssas.exe PID change without a full system restart? Is it safe to read once and cache the value.

VihasMakwana: From https://en.wikipedia.org/wiki/Local_Security_Authority_Subsystem_Service,

> Forcible termination of lsass.exe will result in the system losing access to any account, including NT AUTHORITY, prompting a restart of the machine

Many articles mentioned similar thing i.e. if lsass.exe shut's down, then system is rebooted. 

@gabriellandau what do you think?

gabriellandau: > @gabriellandau Can the lssas.exe PID change without a full system restart?

Negative.  One per boot.  It's [safe to cache](https://github.com/elastic/beats/issues/41407#issue-2609676190).

> Its PID can be found in the registry. We can query that value once and cache the value because it will never change until reboot.

VihasMakwana:
@swiatekm @mauri870 I've added unit test case and fixed some bug. can you take a look?

mauri870 (approved):

ycombinator:
@gabriellandau could you also re-review this PR, please?  Thanks!

gabriellandau (approved):
Thanks for the change @VihasMakwana.

PR #209, opened 2025-03-20
[registry] use per-namespace registry for platform specific metrics

VihasMakwana:
Use per-namespace registry for platform specific metrics as well.

For some reason, I missed this in https://github.com/elastic/elastic-agent-system-metrics/pull/205 and it wasn't get caught  during reviews as well.


mauri870 (approved):
Just tested locally and it works as expected with two receivers running.

PR #221, opened 2025-05-29
perf: avoid unnecessary io when reading rss memory

kruskall:
## What does this PR do?

memstatsreporter only cares about the rss memory

retrieve the memory info instead of filling the process info with ton of io calls

## Why is it important?

memstat reporter is a hot path and it is called quite often

benchmark shows significant improvement:

```
             │   old.txt    │              new.txt               │
             │    sec/op    │   sec/op     vs base               │
MemReport-20   147.63µ ± 4%   22.35µ ± 2%  -84.86% (p=0.002 n=6)

             │    old.txt    │               new.txt               │
             │     B/op      │     B/op      vs base               │
MemReport-20   33.241Ki ± 0%   1.273Ki ± 0%  -96.17% (p=0.002 n=6)

             │   old.txt   │              new.txt              │
             │  allocs/op  │ allocs/op   vs base               │
MemReport-20   247.00 ± 0%   20.00 ± 0%  -91.90% (p=0.002 n=6)
```

Bench:

```go
func BenchmarkMemReport(b *testing.B) {
	r := MemStatsReporter(logp.NewLogger(""), &process.Stats{Hostfs: resolve.NewTestResolver(""), ProcsMap: process.NewProcsTrack()})
	for i := 0; i < b.N; i++ {
		r(monitoring.Full, monitoring.NewKeyValueVisitor(func(s string, i any) {
			//b.Log(fmt.Sprintf("%v=%v", s, i))
		}))
	}
}
```

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/report/report.go
+++ b/report/report.go
@@ -39,6 +39,15 @@ import (
 )
 
 func MemStatsReporter(logger *logp.Logger, processStats *process.Stats) func(monitoring.Mode, monitoring.Visitor) {
+	p := psprocess.Process{
+		Pid: int32(os.Getpid()),
+	}
+
+	ctx := context.Background()
+	if processStats != nil && processStats.Hostfs != nil && processStats.Hostfs.IsSet() {
+		ctx = context.WithValue(context.Background(), common.EnvKey, common.EnvMap{common.HostProcEnvKey: processStats.Hostfs.ResolveHostFS("")})
+	}
+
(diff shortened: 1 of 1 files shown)


mauri870 (approved):

cmacknz (approved):
The test coverage here is pretty thin, but I manually checked that the source of the RSS memory was calculated the same way before and after this change for each major OS.

PR #224, opened 2025-06-09
perf: avoid unnecessray IO in FDUsageReporter

kruskall:
## What does this PR do?

similar to memstatreporter, we don't need to retrieve all process information.
fdusagereporter only cares about fd limits so only retrieve what we need using gopsutil

## Why is it important?

fdusagereporter is a hot path and it's called quite often

benchmark shows significant improvement

```
              │  old-fd.txt  │             new-fd.txt             │
              │    sec/op    │   sec/op     vs base               │
FDreporter-20   126.33µ ± 3%   35.17µ ± 1%  -72.16% (p=0.002 n=6)

              │  old-fd.txt   │             new-fd.txt              │
              │     B/op      │     B/op      vs base               │
FDreporter-20   33.188Ki ± 0%   9.750Ki ± 0%  -70.62% (p=0.002 n=6)

              │ old-fd.txt │            new-fd.txt             │
              │ allocs/op  │ allocs/op   vs base               │
FDreporter-20   246.0 ± 0%   110.0 ± 0%  -55.28% (p=0.002 n=6)
```

Bench:

```
func BenchmarkFDreporter(b *testing.B) {
	r := FDUsageReporter(logp.NewLogger(""), &process.Stats{Hostfs: resolve.NewTestResolver(""), ProcsMap: process.NewProcsTrack()})
	for i := 0; i < b.N; i++ {
		r(monitoring.Full, monitoring.NewKeyValueVisitor(func(s string, i any) {
			//b.Log(fmt.Sprintf("%v=%v", s, i))
		}))
	}
}
```

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/report/metrics_file_descriptors.go
+++ b/report/metrics_file_descriptors.go
@@ -22,7 +22,6 @@ package report
 
 import (
 	"context"
-	"os"
 
 	"github.com/shirou/gopsutil/v4/common"
 	psprocess "github.com/shirou/gopsutil/v4/process"


report/metrics_file_descriptors.go
```diff
@@ -31,29 +37,45 @@ func SetupLinuxBSDFDMetrics(logger *logp.Logger, reg *monitoring.Registry, proce
 }
 
 func FDUsageReporter(logger *logp.Logger, processStats *process.Stats) func(_ monitoring.Mode, V monitoring.Visitor) {
+	p := psprocess.Process{
+		Pid: int32(os.Getpid()),
+	}
```
AndersonQ, line 42: [Question]

If I understand it correctly, right now, those metrics are always from the process itself. However depending on hostFS it might be the PID the container sees or the host.
Therefore I wonder if using `os.Getpid()` might get the wrong PID

AndersonQ: I don't know, I learnt we have it while working with the beats and k8s/docker.
I know it's used in k8s to avoid giving elevated permissions to a container, so some host paths are mounted on the container, that way the container sees the host but without requiring permissions to do so.
But it's a good question, let me try to dig and see if I find an answer

report/metrics_file_descriptors.go
```diff
@@ -31,29 +37,49 @@ func SetupLinuxBSDFDMetrics(logger *logp.Logger, reg *monitoring.Registry, proce
 }
 
 func FDUsageReporter(logger *logp.Logger, processStats *process.Stats) func(_ monitoring.Mode, V monitoring.Visitor) {
+	pid, err := process.GetSelfPid(processStats.Hostfs)
+	if err != nil {
+		logger.Error("Error while retrieving pid: %v", err)
+		return nil
```
AndersonQ, line 43: I don't think it's safe to return `nil` here. It'll end up being called latter on without a nil check.

Originally there was always a non-nil return as everything would happen on the callback, on error, the callback would log and do nothing. 

I'm not super familiar with how those metrics are used, so I'm not sure if this callback is called only once per application run or if it might be called several times. If it's the latter, I'm wondering if there could be a situation where `process.GetSelfPid` could eventually succeed.
It's just reading a file, it's unlikely the permissions for the file would change in runtime, the process wouldn't get more permissions in runtime. SO I believe handling it before creating the callback might be ok. However it should always return a valid callback, no-op on error.

Let me try to find someone who knows better this part of this lib

VihasMakwana: yeah. It doesn't look safe.

Here's my understanding of this part of the lib:
1. We initialise all the registries in beats.
2. If monitoring is enabled, we log the internal metrics for each registry every 30s by default.
3. In this case, we have setup a callback here https://github.com/elastic/elastic-agent-system-metrics/blob/c33dfcd19bc9991336bbb406f5889bbc99ccb822/report/metrics_file_descriptors.go#L35-L37
   - `FDUsageReporter` returns a function, which is the one responsible for capturing metrics. It will be called when "visit" this registry key. [Here's](https://github.com/elastic/elastic-agent-libs/blob/b06f9b2ef69e8d4007e4113af4dea928450b7646/monitoring/metrics.go#L235) the code responsible for calling it 
   - As you can see, there is no null check in place and I suspect it will create a panic if we return `nil`.
 
 
I suggest we return an empty function 

```suggestion
		return return func(_ monitoring.Mode, V monitoring.Visitor) {
		      	 V.OnRegistryStart()
		          V.OnRegistryFinished()
		}
```

VihasMakwana (changes requested):

report/metrics_file_descriptors.go
VihasMakwana: > I'm not super familiar with how those metrics are used, so I'm not sure if this callback is called only once per application run or if it might be called several times

It will be called as per configured period. The key `logging.metrics.period` in beat configuration is the one that controls the frequency I guess.

VihasMakwana (approved):

AndersonQ (dismissed):

VihasMakwana (approved):

AndersonQ (approved):

PR #234, opened 2025-06-13
fix: turn linux boottime into an atomic var

kruskall:
## What does this PR do?

replace uint with atomic.uint to avoid race condition

## Why is it important?

tests with race detector will fail

## Checklist

## Author's Checklist

- [ ]

## Related issues

discovered by https://github.com/elastic/elastic-agent-system-metrics/pull/232

### DIFF
--- a/metric/system/process/process_linux_common.go
+++ b/metric/system/process/process_linux_common.go
@@ -28,6 +28,7 @@ import (
 	"os/user"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


leehinman (commented):
I agree this will fix the race condition.

I do wonder if `sync.Once` would be a better fit though.  It seems like we really only want to store the value once.  Is there something that is making the lower level atomic a better choice?

pierrehilbert:
@kruskall are you still on top of this one?

PR #259, opened 2025-08-18
Remove global logger usage 

khushijain21:
## What does this PR do?
This PR introduces new methods that accepts a logger. This is done to get rid of global loggers wherever they are used
This is required for beatreceivers to work.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Relates https://github.com/elastic/ingest-dev/issues/5251

### DIFF
--- a/metric/system/cpu/cpu.go
+++ b/metric/system/cpu/cpu.go
@@ -31,13 +31,8 @@ import (
 // Load returns CPU load information for the previous 1, 5, and 15 minute
 // periods.
 // Deprecated: use LoadWithLogger
-func Load(logger *logp.Logger) (*LoadMetrics, error) {
-	avg, err := load.Avg()
-	if err != nil {
-		return nil, err
-	}
-
-	return &LoadMetrics{avg, logp.NewLogger("")}, nil
+func Load() (*LoadMetrics, error) {
+	return LoadWithLogger(logp.NewLogger(""))
 }
 
 // LoadWithLogger returns CPU load information for the previous 1, 5, and 15 minute


khushijain21:
https://github.com/elastic/beats/pull/46054. This PR will bump accommodate it

report/setup.go
```diff
@@ -72,12 +76,12 @@ func SetupMetricsOptions(opts MetricOptions) error {
 		return fmt.Errorf("failed to init process stats for agent: %w", err)
 	}
 
-	monitoring.NewFunc(opts.ProcessMetrics, "memstats", MemStatsReporter(opts.Logger, processStats), monitoring.Report)
-	monitoring.NewFunc(opts.ProcessMetrics, "cpu", InstanceCPUReporter(opts.Logger, processStats), monitoring.Report)
+	monitoring.NewFunc(opts.ProcessMetrics, "memstats", MemStatsReporter(log, processStats), monitoring.Report)
```
khushijain21, line 79: FYI, @kruskall, the previous PR did not check if the logger was nil/unset

metric/system/cgroup/cgstats.go
```diff
@@ -73,7 +74,7 @@ func (stat *StatsV1) FillPercentages(prev CGStats, curTime, prevTime time.Time)
 	if len(stat.CPUAccounting.UsagePerCPU) > 0 {
 		cpuCount = len(stat.CPUAccounting.UsagePerCPU)
 	} else {
-		cpuCount = numcpu.NumCPU()
+		cpuCount = numcpu.NumCPU(logp.NewLogger(""))
```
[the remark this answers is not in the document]
khushijain21: yes, `statsV1` is not used by beats as of now. So we are doing breaking changes only where necessary and using global loggers everywhere else

AndersonQ:
> This is a breaking change so I'd rather not approve but left a review

It's a good point. Even though we don't do any compatibility promises, we tend to avoid breaking changes. 
On the other hand, looking at pkg.go.dev ([here](https://pkg.go.dev/github.com/elastic/elastic-agent-system-metrics/metric?tab=importedby) and [here](https://pkg.go.dev/github.com/elastic/elastic-agent-system-metrics@v0.13.0/report?tab=importedby)), it seems only beats/agent or forks of it use the lib.

A simple approach is to deprecate the exiting functions and create new ones receiving a logger. It's a bit more work, but it accomplishes the goal without a breaking change.
For constructor like functions using functional options can be a way to introduce the logger with a backward compatible signature change.

khushijain21:
>IMO adding a (deprecated) numcpu func with no arg that defaults to calling the new func with nooplogger should be good enough

Well there is `cpu.Load()` as well that accepts a logger. If there is a consensus that we want to do it this way every time then I'll change it.

metric/system/diskio/diskstat_linux.go
```diff
@@ -39,7 +40,7 @@ func GetCLKTCK() uint32 {
 }
 
 // IOCounters should map functionality to disk package for linux os.
-func IOCounters(names ...string) (map[string]disk.IOCountersStat, error) {
+func IOCounters(_ *logp.Logger, names ...string) (map[string]disk.IOCountersStat, error) {
```
VihasMakwana, line 43: we can skip this part if we don't require a logger. AFAIK, `disk.IOCounters` is a part of `gopsuitl` and we won't pass logger to it.

metric/system/diskio/diskstat_other.go
```diff
@@ -47,6 +49,6 @@ func (stat *IOStat) CalcIOStatistics(rcounter disk.IOCountersStat) (IOMetric, er
 func (stat *IOStat) CloseSampling() {}
 
 // IOCounters should map functionality to disk package for linux os.
-func IOCounters(names ...string) (map[string]disk.IOCountersStat, error) {
+func IOCounters(_ *logp.Logger, names ...string) (map[string]disk.IOCountersStat, error) {
```
VihasMakwana, line 52: Same as above. We can skip for now.

VihasMakwana (commented):
Some comments

metric/system/diskio/diskstat_linux.go
VihasMakwana: got it.

PR #277, opened 2025-12-22
Fetch per process swap memory data (linux)

stefans-elastic:
## What does this PR do?

Add collection of data on how much swap memory monitored process is using. This is achieved by parsing of `/proc/[pid]/status` and getting value of `VmSwap` field. The value in the field is in KB units so we also convert original value to bytes.
This is linux implementation. Implementations for windows and macos will be implemented in separate PRs.

## Why is it important?

This change is needed because we've received a customer request to add this feature.

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/metric/system/process/process_linux_common.go
+++ b/metric/system/process/process_linux_common.go
@@ -550,9 +550,22 @@ func getSwapData(hostfs resolve.Resolver, pid int) (opt.Uint, error) {
 		return opt.Uint{}, fmt.Errorf("error getting process status(PID %d): %w", pid, err)
 	}
 
-	if swapKb, ok := statusMap["VmSwap"]; ok {
-		return opt.UintWith(swapKb << 10), nil // the bit shift converts value from kB to bytes
+	if swapStr, ok := statusMap["VmSwap"]; ok {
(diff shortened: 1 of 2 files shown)


metric/system/process/process_linux_common.go
```diff
@@ -538,3 +544,41 @@ func getProcState(b byte) PidState {
 func FillMetricsRequiringMoreAccess(_ int, state ProcState) (ProcState, error) {
 	return state, nil
 }
+
+func getSwapData(hostfs resolve.Resolver, pid int) (opt.Uint, error) {
+	path := hostfs.Join("proc", strconv.Itoa(pid), "status")
+	data, err := os.ReadFile(path)
```
andrzej-stencel: Did you consider using the `getProcStatus` function? It parses the whole `/proc/{pid}/status` file and returns a map. That would be probably easier to read and maintain, although less performant.

andrzej-stencel (approved):

andrzej-stencel (approved):
Thanks Stefan! Looks good to me.

PR #282, opened 2026-02-10
Always use the local filesystem for self-monitoring metrics

belimawr:
## What does this PR do?
Always use the local filesystem for self-monitoring metrics.
Metricbeat will use the hostfs to set the environment variables:
- HOST_PROC
- HOST_SYS
- HOST_ETC

which are read by github.com/shirou/gopsutil/v4 and used when fetching metrics. However, the hostfs is not available by the time self monitoring metrics are initialised, so when running in a container environment, we cannot resolve the PID relative to the hostfs PID namespace.

This causes `elastic-agent-system-metrics` to get the PID using `os.Getpid`, and `gopsutil` uses the hostfs to fetch all metrics, causing us to get metrics from the wrong process or getting an error because `/hostfs/proc/<PID>/fd` does not exist.

Always using the local filesystem for self-monitoring metrics solves this problem.

## Why is it important?
It allows a Beat to correctly get metrics from its own process.

## Checklist

~~## Author's Checklist~~

## Related issues
- https://github.com/elastic/beats/issues/48690

### DIFF
--- a/report/setup.go
+++ b/report/setup.go
@@ -177,7 +177,7 @@ func setupPlatformSpecificMetrics(logger *logp.Logger, processStats *process.Sta
 type localProcResolver struct{}
 
 func (t localProcResolver) ResolveHostFS(path string) string {
-	return filepath.Join(filepath.Separator, path)
+	return filepath.Join("/", path)
 }
 
 func (t localProcResolver) Join(path ...string) string {


report/report.go
```diff
@@ -49,7 +49,7 @@ func MemStatsReporter(logger *logp.Logger, processStats *process.Stats) func(mon
 
 	ctx := context.Background()
 	if processStats != nil && processStats.Hostfs != nil && processStats.Hostfs.IsSet() {
-		ctx = context.WithValue(context.Background(), common.EnvKey, common.EnvMap{common.HostProcEnvKey: processStats.Hostfs.ResolveHostFS("")})
+		ctx = context.WithValue(context.Background(), common.EnvKey, common.EnvMap{common.HostProcEnvKey: processStats.Hostfs.ResolveHostFS("/proc")})
```
AndersonQ: To make sure I understand, you're setting `/proc` because `HOST_PROC` isn't set, is it? Any chance this code runs on windows? It can compile to windows

belimawr: So, the code you're commenting on is generic (all OSes), however the part that uses the context is OS specific.

This context is used by https://github.com/elastic/elastic-agent-system-metrics/blob/243a49474207dfc5a8de1c57abff19e90f1ae630/report/report.go#L69

`MemoryInfoWithContext` is OS specific and part of `github.com/shirou/gopsutil`.

The same is true for `InstanceCPUReporter` from https://github.com/elastic/elastic-agent-system-metrics/blob/243a49474207dfc5a8de1c57abff19e90f1ae630/report/report.go#L79

The context is only used in a Linux specific implementation.


You can also see the test passing on [Windows](https://buildkite.com/elastic/elastic-agent-system-metrics/builds/926#019c521f-fbee-44e4-9bf7-9f4caea8e78e/L296) and [MacOS](https://buildkite.com/elastic/elastic-agent-system-metrics/builds/926#019c521f-fbef-430a-9870-9ebc0c78c29c/L248).

PR #294, opened 2026-05-20
[cgv2] Convert cpu.stat microseconds to nanoseconds at parse site

orestisfl:
## What does this PR do?

In `metric/system/cgroup/cgv2/cpu.go`, `getStats` reads `cpu.stat` which reports CPU time in microseconds (`usage_usec`, `user_usec`, `system_usec`) and previously assigned those values directly to fields named `*.NS`, with no µs -> ns conversion. The cgroups v1 reader in `cgv1/cpuacct.go` does normalize to nanoseconds (`convertJiffiesToNanos`), so v1 was already correct -- only v2 was wrong.

This PR:

- Scales `usage_usec`, `user_usec`, and `system_usec` by `uint64(time.Microsecond)` at the parse site, matching the time-based unit conversion already used in `cgv1/cpuacct.go`. `throttled_usec` is intentionally left in microseconds (its destination field is `.Us`).
- Adds `TestGetStatsUsecToNanos`, a focused regression test that pins the µs -> ns contract with small, hand-picked values.
- Updates `TestGetCPU` expectations that had encoded the bug as truth.

## Why is it important?

The v2 `cpu.stats.{usage,user,system}.ns` values were 1000x too small. Downstream, `cgstats.go:FillPercentages` divides this ``NS`` delta by a true nanosecond wallclock delta, so the derived `cpu.stats.{usage,user,system}.{pct,norm.pct}` were also off by 1000x.

`metric.Round` truncates to 4 decimal places (`DefaultDecimalPlacesCount = 4`), so for realistic workloads the buggy values rounded to literal `0.0000`:

| Real CPU load (1 core) | Real ``pct`` | Buggy ``pct`` (÷1000) | After ``Round`` (4dp) |
|---|---|---|---|
| 100% | 1.0 | 0.001 | 0.001 |
| 50% | 0.5 | 0.0005 | 0.0005 |
| 10% | 0.1 | 0.0001 | 0.0001 |
| 5% | 0.05 | 0.00005 | 0.0001 |
| 1% | 0.01 | 0.00001 | **0.0000** |

``norm.pct`` is further divided by ``cpuCount``, so on multi-core hosts almost everything below saturation lands at ``0.0000``.

## Risk

Behavior change for any consumer that compensated for the bug (e.g. multiplying the v2 ``ns`` field or ``pct`` by 1000). Worth calling out in release notes downstream.

## Checklist

## Related issues

- Closes #293

### DIFF
--- a/metric/system/cgroup/cgv2/cpu.go
+++ b/metric/system/cgroup/cgv2/cpu.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/elastic/elastic-agent-libs/opt"
 	"github.com/elastic/elastic-agent-system-metrics/metric/system/cgroup/cgcommon"
@@ -131,12 +132,14 @@ func getStats(path string) (CPUStats, error) {
 			return data, fmt.Errorf("error parsing cpu.stat file: %w", err)
 		}
 		switch key {
+		// cpu.stat reports CPU time in microseconds; the destination *.NS
+		// fields are nanoseconds (matching cgroups v1), so scale at parse.
 		case "usage_usec":
-			data.Usage.NS = val
+			data.Usage.NS = val * uint64(time.Microsecond)
 		case "user_usec":
-			data.User.NS = val
+			data.User.NS = val * uint64(time.Microsecond)
 		case "system_usec":
-			data.System.NS = val
+			data.System.NS = val * uint64(time.Microsecond)
 		case "nr_periods":
 			data.Periods = opt.UintWith(val)
 		case "nr_throttled":
--- a/metric/system/cgroup/cgv2/v2_test.go
+++ b/metric/system/cgroup/cgv2/v2_test.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


VihasMakwana (approved):

belimawr (approved):
@orestisfl could you check if those fields are used by integrations? I believe the system integration uses them, if so, they might be compensating for this bug.

We need to make sure we're not breaking any integration/dashboard with this release.

orestisfl:
> @orestisfl could you check if those fields are used by integrations? I believe the system integration uses them, if so, they might be compensating for this bug.
> 
> We need to make sure we're not breaking any integration/dashboard with this release.

@belimawr Good call. I checked the system integration, the wider integrations repo, beats, and cloud. TL;DR: nothing compensates for the bug, and the field descriptions already say what the fix produces, so no downstream changes are needed for this PR to land safely.

Affected fields:
- `system.process.cgroup.cpu.stats.{usage,user,system}.ns`
- `system.process.cgroup.cpu.stats.{usage,user,system}.{pct,norm.pct}`

### `elastic/integrations`

- `packages/system/data_stream/process/fields/fields.yml` declares them with units that already match the fix
- No Kibana dashboard, alerting rule template, saved search, or ingest pipeline in any package references these fields.

### `elastic/beats`

Field descriptions in `metricbeat/module/system/process/_meta/fields.yml` have always claimed nanoseconds.

No Go code in beats references the affected paths or applies any compensation.

### `elastic/cloud`

Performance dashboards/wikis use `cgroup.cpuacct.*` (v1), `cgroup.cpu.cfs.{quota,period}.us`, and `cgroup.cpu.stats.throttled.{ns,periods}`/`stats.periods`. None of the fields fixed by this PR appear anywhere.

